PR #59, opened 2022-10-20
Adding 'copystyle' command

pcagas:
Adding `copystyle` command which allows you to make a copy of the default Postgkyl Matplotlib style file, modify it, and then use it in the `plot` command using the `--style FILENAME` option


ammarhakim:
I am merging this PR. However, perhaps we should add options to the command to mod the various params so the style-file does not need to be edited and all options can be set from CLI itself

PR #62, opened 2022-11-23
Fixing bug in select which was picking slightly shifted indices

pcagas:
The issue with `select` picking the wrong indices discovered by @manauref has been fixed


pcagas:
`pgkyl ctest_proj_maxwellian_on_basis_test_1x1v_p1.gkyl -t in interp -p1 -b ms sel -u in -t plus --z1 4e-4 sel -u in -t minus --z1 -4.e-4 pl -u plus,minus -f0`

image

manauref:
I'll just comment that this is very much appreciated, and seems to work. It also helps me understand some asymmetries I was seeing.

PR #69, opened 2023-04-04
fixing types for latest numpy

viper2642:
np.int and np.float are deprecated since numpy 1.16 (if I am not mistaken)

### DIFF
--- a/postgkyl/data/dg.py
+++ b/postgkyl/data/dg.py
@@ -309,9 +309,9 @@ def _interpOnMesh(cMat, qIn, nInterpIn, basisType, c2p=False):
     numInterp[-1] = nInterpIn+1
   #end
   if c2p:
-    qOut = np.zeros(numCells*(numInterp-1)+1, np.float)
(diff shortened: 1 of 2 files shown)


gwhammett:
Interesting.  This reminds me that when teaching about floating point issues a couple of years ago, I discovered that Python's float and Numpy's float64 are not exactly equivalent.  The numpy.float64 type follows IEEE floating point standards, and one can turn off run-time fatal error messages caused by divide by zero, which is sometimes useful and assumed in some numerical algorithms.  This allows one to have r = a/z to give Inf if z=0, and then use 1/r later to get 0.  This is useful in calculating the net resistance of 2 resistors in parallel for example:
rnet = 1.0/(1.0/r1 + 1.0/r2)

I discuss this some on p. 6-8 of:
https://drive.google.com/drive/u/1/folders/1IUtL2lPCTBVRprmmb00H548-14b0eAf5
and related info in the Lecture 5 folder of:
https://drive.google.com/drive/folders/1yYyuc4gvSuB2uF7OetzmYE2QtwLuUlgr?usp=share_link

PR #97, opened 2023-08-22
Using adios2 python binding for reading adios data

liangwang0734:
Motivation:
1. Avoid numpy errors due to unmaintained adios v1 python code
2. Stay with the newer, maintained upstream code
3. Allow possible switch to adios2 I/O in Gkyl in the future

Tests:
- Tested with Python 3.11 and adios2 2.9.1 (from [conda-forge](https://anaconda.org/conda-forge/adios2))
  - I created a new conda env to make the dependency resolution easier
- Tested on gkyl g2 fluid moment FV data, and `fieldEnergy` data;
- NOT tested for DG data or c2p.

Other notes:
- Commented-out codes were not adapted

Updates:
- [x] the `nstep=1` feature tested with @pcagas case
  - `pgkyl test.bp -d default_9 pl`
- [x] partial loading with `pgkyl Regression/mom-gem-challenge/rt-10m-gem_field_0.bp sel --z0 1:32 pl`
- [x] 2d DG data from `rt-weibel-2x2v-p1.lua`
- [x] `c2p` test using `pgkyl ls2-lapd-3x2v-p1_elc_gridDiagnostics_0.bp -d Temp --c2p ls2-lapd-3x2v-p1_mapc2p.bp interp -b ms sel --z2 10 pl -a`
- [x] `write` feature updated the corresponding unit test `test_WriteBp` 
- [x] `write` feature tested with `pgkyl ls2-lapd-3x2v-p1_elc_gridDiagnostics_0.bp -d Temp --c2p ls2-lapd-3x2v-p1_mapc2p.bp interp -b ms sel --z2 10 write -f test.bp`
  - the result can be plotted with `pgkyl test.bp -d Temp pl` normally
- [x] `write` feature tested on restart file of `Regression/mom-gem-challenge/rt-5m-gem.lua`; the written restart file seems to be readable by gkyl
  - run `rt-5m-gem.lua` as normal
  - `mv rt-5m-gem_elc_restart.bp rt-5m-gem_elc_restart.old.bp`
  - `pgkyl rt-5m-gem_elc_restart.old.bp write -f rt-5m-gem_elc_restart.bp`
  - modify `tEnd` and `nFrame` in `rt-5m-gem.lua` 
  - restart the simulation using the new restart file written by postgkyl

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,8 +5,6 @@ dependencies:
   - matplotlib
   - click
   - conda-forge::adios2=2.9.1
-  - conda-forge::adios=1.13.1
-  - conda-forge::adios-python=1.13.1=py38h522c649_1006
   - pytables
   - scipy
   - sympy
--- a/environment.yml
+++ b/environment.yml
@@ -4,6 +4,7 @@ dependencies:
   - numpy=1.21.*
   - matplotlib
   - click
+  - conda-forge::adios2=2.9.1
   - conda-forge::adios=1.13.1
   - conda-forge::adios-python=1.13.1=py38h522c649_1006
   - pytables
--- a/meta.yaml
+++ b/meta.yaml
@@ -32,8 +32,6 @@ requirements:
     - click
     - sympy
     - conda-forge::adios2=2.9.1
-    - conda-forge::adios=1.13.1
-    - conda-forge::adios-python=1.13.1=py38h522c649_1006
     - adios
     - adios-python
     - msgpack-python
--- a/meta.yaml
+++ b/meta.yaml
@@ -31,6 +31,7 @@ requirements:
     - pytables
     - click
     - sympy
+    - conda-forge::adios2=2.9.1
     - conda-forge::adios=1.13.1
     - conda-forge::adios-python=1.13.1=py38h522c649_1006
     - adios
--- a/meta.yaml
+++ b/meta.yaml
@@ -32,8 +32,6 @@ requirements:
     - click
     - sympy
     - conda-forge::adios2=2.9.1
-    - adios
-    - adios-python
     - msgpack-python
 
 test:
--- a/postgkyl/data/gdata.py
+++ b/postgkyl/data/gdata.py
@@ -174,7 +174,7 @@ class GData(object):
 
   def getInputFile(self):
     import adios2
-    fh = adios2.open(self.file_name, 'r')
+    fh = adios2.open(self.file_name, 'rra')
     inputFile = fh.read_attribute_string('inputfile')[0]
     fh.close()
     return inputFile
--- a/postgkyl/data/gdata.py
+++ b/postgkyl/data/gdata.py
@@ -381,6 +381,9 @@ class GData(object):
       var_name = self._var_name
     #end
 
+    values = np.empty_like(self.getValues())
+    values[...] = self.getValues()
+
     if mode == 'bp':
 
       import adios2
@@ -397,14 +400,14 @@ class GData(object):
           fh.write('time', self.ctx['time'])
         #end
 
-        fh.write(var_name, self.getValues(), full_shape, offset, full_shape)
+        fh.write(var_name, values, full_shape, offset, full_shape)
 
         fh.close()
 
       else:
 
         fh = adios2.open(out_name, "a")
-        fh.write(var_name, self.getValues(), full_shape, offset, full_shape)
+        fh.write(var_name, values, full_shape, offset, full_shape)
         fh.close()
 
       #end
@@ -415,7 +418,6 @@ class GData(object):
       for d in range(num_dims):
         grid[d] = 0.5*(grid[d][1:]+grid[d][:-1])
       #end
-      values = self.getValues()
 
       basis = np.full(num_dims, 1.0)
       for d in range(num_dims-1):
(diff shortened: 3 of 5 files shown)


pcagas:
This is looking quite good! There are a few things that need to be addressed before we can merge:
- Please do test DG and c2p data.
- Remove the adios1 dependencies.
- Does the partial loading work?
- The `nsteps=1` in `data = var.read(offset=offset, count=count, nsteps=1)` is important for reading some data. Can we add it in adios2?
- pgkyl can also write bp files. This needs to be updated to adios2 as well. It's in the `write` function in the `gdata.py`

pcagas (changes requested):
This is looking quite good! There are a few things that need to be addressed before we can merge:
- Please do test DG and c2p data.
- Remove the adios1 dependencies.
- Does the partial loading work?
- The `nsteps=1` in `data = var.read(offset=offset, count=count, nsteps=1)` is important for reading some data. Can we add it in adios2?
- pgkyl can also write bp files. This needs to be updated to adios2 as well. It's in the `write` function in the `gdata.py`

liangwang0734:
Hi @pcagas , I can do the writing relatively quickly. But I'm not optimistic that it will write files readable by Gkyl for restarting, etc. How are the files written by postgkyl supposed to be used?

liangwang0734:
Also, what data needs the `nstep=1` feature of adios v1? Is there something I can test against?

ammarhakim:
We do not use output from pgkyl for restarts. Mostly, we use it for doing more down-stream analysis. So output from a complex chain of commands could be stored and reused in the next invovation of pgkyl. It is also useful when converting from other formats. For example, reading h5 and saving as BP.

I think we should be very cautious of merging this PR. We should take this opportunity to do extensive tests. As pgkyl and BP are so central we do not wish to break any existing functionality and so the PR should only be accepted after extensive testing.

JunoRavin:
The biggest utility of the write method for me is for down-selection and data transfers. When we do kinetic simulations with 100 GB+ distribution function outputs, we use pgkyl to down select and write out a chunk of that distribution function, then transfer that data off the cluster for more streamlined analysis on our personal machines.

pcagas:
Currently, the bp data that postgkyl writes _can_ be used restarts. I have used it several times.

liangwang0734:
It would be good to have more tests.

Most of tests in the unit folder is broken (for minor reasons, though).

Jimmy, can you give a small sample data of the input and extraction output
from the write feature?

pcagas:
Currently, the bp data that postgkyl writes _can_ be used restarts. I have used it several times.

pcagas:
The `nstep` was needed to read back some data that pgkyl writes. If I remember correctly, this was a feature where we packed several variables into one bp file. I think some Jimmy's colleagues wanted that

liangwang0734:
> This doesn't seem to be working for me. With adios2 2.8, I can plot everything except for `dynVector` data. I have tried updating to 2.9 and that replaced a lot of mainline packages with `conda-forge` ones. Now I'm getting:
> 
> `python: symbol lookup error: /home/cagas65/conda/envs/py311/lib/python3.11/site-packages/mkl/../../../libmkl_intel_thread.so.2: undefined symbol: omp_get_num_procs`

It seems an MPI problem. The adios 2 python library can be built with MPI and possibly there is some mismatch here.

What's the full error, and what's the exact version of the adios 2.9 you have? One solution might be to force to use an non-mpi version of adios 2.9? Mine looks like (grep'ed from `conda list`):

```
adios2                    2.9.1           nompi_py311hc31d2c9_101    conda-forge
```

pcagas:
Oh, wait... it's something different. The adios is working but the update to the `conda-forge` `adios` 2.9 messed up my plotting.
```
$ pgkyl rt-two-stream-p2_elc_0.bp info
(default#0)
├─ Time: 0.000000e+00
├─ Frame: 0
├─ Number of components: 8
├─ Number of dimensions: 2
├─ Grid: (uniform)
│  ├─ Dim 0: Num. cells: 64; Lower: -6.283185e+00; Upper: 6.283185e+00
│  └─ Dim 1: Num. cells: 32; Lower: -6.000000e+00; Upper: 6.000000e+00
├─ Maximum: 1.676015e+00 at (31, 18) component 0
├─ Minimum: -4.698334e-01 at (32, 19) component 2
├─ DG info:
│  ├─ Polynomial Order: 2
│  └─ Basis Type: serendipity (modal)
├─ Created with Gkeyll:
│  ├─ Changeset: 9663ea594e80
│  └─ Build Date: Jun 14 2023 10:04:32
├─ charge: -1
├─ mass: 1
├─ grid_type: uniform
```
This works!

But this is a problem now...
```
$ pgkyl rt-two-stream-p2_fieldEnergy.bp pl
python: symbol lookup error: /home/cagas65/conda/envs/py311/lib/python3.11/site-packages/mkl/../../../libmkl_intel_thread.so.2: undefined symbol: omp_get_num_procs
```

pcagas (approved):
I have resolved the MPI problem. This is what happened: For some reason, when I first installed the dependencies I only got `adios2` 2.8 which doesn't work with our `dynVectors`. Running `conda update -c conda-forge adios` updated it to 2.9 but also replaced many of my packages with `conda-forge` versions and (I think) resulted in the `mkl` bug. So I have tried running `conda update `mkl` which replaced many of the packages back with the mainline versions while keeping the `adios2` 2.9!

Everything seems to be working and even the hustle explained above is better than what we need to do now with `adios1` and `numpy`.

PR #105, opened 2023-09-15
More minor fix for the adios2

liangwang0734:
- fix the `var_name` check that sets the `is_frame` flag
- other minor changes to make the code more robust; see the commits

These are perhaps not needed usually in the command line mode, but I got an error when I explicitly specify the `reader_name` to `adios` in the script mode since the relevant flags (`is_frame` etc. are not set).

I believe other readers might have this problem (i.e., when the reader name is explicitly specified, some flags might not be set when used in the reading function).

### DIFF
--- a/postgkyl/data/gdata.py
+++ b/postgkyl/data/gdata.py
@@ -94,6 +94,9 @@ class GData(object):
     file_name = str(file_name)
     self.file_name = file_name
     self.mapc2p_name = mapc2p_name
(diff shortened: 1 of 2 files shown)


pcagas (approved):
Yup, thanks for catching that. I have slightly changed the fix so that the reader is now checked every time even when manually specified. This will then robustly apply to all the readers.

PR #107, opened 2023-10-19
Changing the adios2 read mode

pcagas:
Changing the adios read mode everywhere from `r` to `rra`. This is connected to https://github.com/ornladios/ADIOS2/issues/3843

### DIFF
--- a/postgkyl/data/read_gkyl_adios.py
+++ b/postgkyl/data/read_gkyl_adios.py
@@ -34,7 +34,7 @@ class Read_gkyl_adios(object):
     # imported when actially needed
     try:
       import adios2
-      fh = adios2.open(self.file_name, "rra")
+      fh = adios2.open(self.file_name, 'rra')
       if self.var_name in fh.available_variables():
         self.is_frame = True
       for key in fh.available_variables():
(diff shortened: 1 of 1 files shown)


postgkyl/data/read_gkyl_adios.py
```diff
@@ -34,7 +34,7 @@ def _is_compatible(self) -> bool:
     # imported when actially needed
     try:
       import adios2
-      fh = adios2.open(self.file_name, "rra")
+      fh = adios2.open(self.file_name, 'rra')
```
manauref, line 37: @pcagas 
I assume we don't want to support "TimeMesh" and "Data", without a number after these names, right? (i.e. We only want to support TimeMesh0/Data0, TimeMesh1/Data1, etc)

Because presently gkyl is setup to produce such files but maybe it thouldn't (?). Maybe that's legacy and we should remove it from gkyl

postgkyl/data/read_gkyl_adios.py
pcagas, line 1: @manauref I really hope this doesn't bite us in the future, but I don't see a reason for changing this check (line 41) from its original form. I'm reverting it back to where it was and now it should simply work for all the cases.

pcagas, line 1: @manauref I really hope this doesn't bite us in the future, but I don't see a reason for changing this check (line 41) from its original form. I'm reverting it back to where it was and now it should simply work for all the cases.

pcagas: Also, how did you directly add some diffs to the comment? that's really helpful

manauref: Not sure. I think i went to Files changed, and I reviewed the code, so my comments appeared here with the corresponding code/diff

Issue #27, opened 2021-04-20
List command for files with several datasets/varnames


nmandell:
We have some output files which contain several datasets, each with their own varname. For example, in GK, we have an `*_allGeo_0.bp` file, which contains several geometry-related quantities. A particular quantity can be accessed using the `-d` flag, e.g. `pgkyl '*_allGeo_0.bp' -d bmag info`. 

Perhaps the `listoutputs` command can be extended for this purpose, so that it will show all of the available varnames/datasets in such a file. This could be used as
```
> pgkyl '*_allGeo_0.bp' list
bdriftX 
bmag    
cmag    
bmagInv 
bdriftY 
gxx     
gxy     
gyy
```

manauref:
Alternatively this could be folded into `pgkyl info` so that `pgkyl
sim_allGeo_0.bp info` without the `-d dataset` simply lists all the
datasets? not sure. We also need to keep in mind how these things will work
when we have multiple blocks and we have things like

block1/bmag
block1/bdriftY
...
block2/bmag
block2/bdriftY
...
etc

manauref:
Also, I forgot to mention that this can be accomplished with `bpls`
```
bpls <simName>_allGeo_0.bp
```

pcagas:
At this point, specifying a file in the postgkyl chain loads the file (this might change in the future). So this functionality of listing varnames after the file is loaded doesn't currently make much sense to me. I can easily create a new command that lists varnames for a specified file but that duplicates the functionality of `bpls` that @manauref mentioned.

I think it is currently cleaner to keep it as it is but I can be convinced otherwise.

ammarhakim:
This old issue was closed without any resolution. The multi-data file are not readable from pgkyl and I can't find the list of data-sets in the file. Using bpls is not an option as we can't rely on any external tools like that. What happens if the multi-data files are say native G0 files in the future, or Flash files?

What I recommend is we extend the info command to handle this. Info could list all files and print further info on each of them. This would remove the dependency on bpls which may not be installed and anyway the file may not be a BP file to start with.

The fact that pgkyl barfs on Gkeyll generated output is not a good look for us. We should never fail on any BP or Gkyl file we ourselves generate. Even if we just handle info for some files that is enough.

pcagas:
A draft with a solution to this issue in the `loading_revamp` revamp branch (also currently a PR draft #111).

[...]

PR #111, opened 2024-01-08
Loading revamp

pcagas:


pcagas:
Connected to the issue #27

PR #119, opened 2024-02-28
Update of the installation system

pcagas:
This PR refactors the way postgkyl is installed: using this repository is now the only supported option; conda is still being used for installing dependencies.

The goal of these changes is to increase reliability across several machines/systems and provide a unified experience.

For installation, we recommend removing the current installation of pgkyl, creating a virtual environment, and installing the dependencies
through [conda](https://conda.io/miniconda.html):
```
conda env create -f environment.yaml
```

The environment is then activated with
```
conda activate pgkyl
```

However, one can also attempt to install the dependencies directly to the current
conda environment using:
```
conda install --file requirements.txt
```

With all the dependencies installed, both the library and the command line tool
`pgkyl` can then be installed from the source with `pip`:
```
pip install -e .
```

To deactivate the environment, use
```
conda deactivate
```


pcagas:
Additionally, the first `pytest` tests are added.

Issue #14, opened 2021-01-04
Diagnostic enhancements of postgkyl


JunoRavin:
This issue can possible be split into multiple issues/sprints, but for now collecting the most high priority diagnostic enhancements that would streamline scripting analysis of Vlasov and fluids simulations:

* Extensions to primitive moments (and support for Vlasov simulations)
  ** Infrastructure to check if input is "finite-volume-like" or a DG field. If "finite-volume-like" call routines that compute u, T, P_{ij}, etc. (and add the primitives of the third moment)
  ** If data is a DG field, compute the primitive moments using weak operations (see Issue #13)

* Drift analysis
  ** Compute total and individual components of parallel and perpendicular flows (see attached image for example of vector perpendicular flow)
  ** By default, could use np.gradient for "finite-volume-like" data and differentiate() for a DG field. Long-term, the accuracy of gradient operations could be improved by the DG field being passed to these operations being a DG field computed from recovery. 

* Energization analysis
  ** Using drift analysis, compute total parallel/perpendicular energization, and energization from various components of the flows
  ** Implement additional common energization metrics (J dot E, Zenitani, p-Theta, Pi-D)

* Auxiliary diagnostics
  ** Swisdak measure of agyrotropy

* Field-particle correlation
  ** Total correlation |v|^2 E dot grad_v f
  ** Component-wise correlation v_i E_i df/dv_i
  ** General fixed-time frame-transformed correlation (v_{sim} -> v_{sim} - U, E_{sim} -> E_{sim} + U x B_{sim})
  ** Total parallel correlation ((v dot b_hat) b_hat)^2 * ((E dot b_hat) b_hat) df/d((v dot b_hat) b_hat)
  ** Total perpendicular correlation
  ** General aligned correlation which takes the magnetic field and a flow vector as inputs and computes three new velocity/field variables: v_{parallel} = (v dot b_hat) b_hat, v_{perp_2} = (U x b_hat) dot v (in direction U x b_hat), and v_{perp_1} = (b_hat x (U x b_hat)) dot v (in direction b_hat x (U x b_hat))
     *** Since many of these are total correlations, could have a single function which takes as input the velocity vector, electric field vector, and distribution function, and computes the instantaneous or averaged correlation from these three quantities
     *** Separate functions could be used to construct the transformed variables (and transformed grid to compute derivatives on)
[...]

PR #140, opened 2024-06-28
Formating and syntax cleanup

pcagas:
This substantive change in terms of additions and deletions aims to improve the readability of the code, unify formatting across the whole package, and add docstrings where they were missing.

**Minimal functional changes are made and all the tests are passing.**

There are several notable changes:

# Change from camelCase to snake_case 

It is common in Python to use the snake case; however, Postgkyl originally used the camel case similar to Gkeyll 1. But with `gkylzero` moving to the snake case, there is no reason to go against Python conventions.

Still, Postgkyl is a Physics-motivated code; therefore, some names go against the convention. For example, moments still use capital `M0` or `M1`, no matter how much `pylint` complains.

# Switch to f-strings

The second stylistic change is the switch to f-strings which are generally easier to read. Following is an example of a classical string:
```python
print("I am adding {:f} and {:f}".format(a, b))
```
and f-string:
```python
print(f"I am adding {a:f} and {b:f}")
```
Note the required `f`.

# Typing and docstrings

Postkyl will be using the typing. It improves code readability for both users and developers and enables dynamic code checking and hinting by an IDE. It also clearly shows which function parameters are optional.

For example:
```python
def laguerre_compose(in_f: GData | Tuple[list, np.ndarray],
    in_T_m: GData | Tuple[list, np.ndarray],
    out_f: GData | None = None) -> Tuple[list, np.ndarray]:
```
The input `in_f` can be either out `GData` calls or a tuple of grid and values (more on this duality in the next section). The same applies to the T/m variable. The output, `out_f`, which allows to write out the result into a GData object is marked as optional by the `| None` option. This syntax also marks that the function will return a grid (the list) and values (the NumPy array).

Following is an example of IDE hinting:
image

In addition to the typing, docstrings have been added to several functions. This is far from finished but it will be an ongoing process.

# Input to functions

It would be generally useful if the library of Postgkyl functions could work both with the internal `GData` class and also simply with NumPy arrays. I think that the best solution is through indirection. As was mentioned above, the Postgkyl function can now take in either a single `GData` object or a tuple of grid and values, `in_f: GData | Tuple[list, np.ndarray]`. This is then followed in the code by unwrapped:
```
in_f_grid, in_f_values = input_parser(in_f)
```
As a result, the rest of the function can be (almost) completely agnostic to the input format.

Consequently, this works both ways and for example, `plot` can finally take a grid and values!
image

This change is quite beneficial for the work on #14


pcagas:
closes #142

PR #147, opened 2024-08-21
Multiblock Plotting/Animation/Simple Lineout Support

lucaa-g:
# Changes to Postgkyl

This PR supports the implementation of plotting, animation, and basic lineouts for Multiblock simulation files in Postgkyl
## Plot Function

1. Added **multiblock** (or mb) parameter
This parameter automatically plots all loaded data objects onto the same figure and sets an appropriate scale. It does so using the ‘figure’ and ‘globalrange’ kwargs.
2. Added **cutoffglobalrange** (cogr) parameter
This allows the user to adjust the limits of the scale by choosing the middle percentile of values to keep inside the scale. It does this by creating an array containing all the extrema from each block and setting the y/z limits to the edges of the middle percentile chosen by the user.
### Small Fix:
I found that there was no way to toggle OFF the colorbar of a 2D plot due to a conditional error in the plot function in the output directory. This caused lag in multiblock animations due to colorbar stacking. I also fixed this.

### Example code:
`pgkyl <blocks to be loaded> plot --multiblock (or -mb)`
- Loaded blocks should belong to one frame

## Command Util

1. Added **set_frame** function
This function uses the file names of all the loaded blocks to find the relative frame for each block, then sets the ‘frame’ value in each block’s ctx dictionary to that frame. This function is used in both animate and select in order to organize and group blocks for animations and lineout animations

## Animate Function

For the animate function to work with multiblock data, I had to readjust its framework. The function creates a list of lists, where each interior list contains all the block data objects which correspond to a specific frame. So, each interior list will contain however many blocks there are per frame, or just 1 block per frame if you are animating a non-multiblock case. The new **multiblock** parameter uses the util **set_frame** function to sort all the loaded data objects, then plot all the blocks per frame and animate it using matplotlib’s FuncAnimation and an altered version of the previous **_update** function. The new **_update** function iterates through every frame list rather than every offset.
### Example code:
`pgkyl <blocks from all frames to be loaded> animate --multiblock`

### Other small fixes
- Only the first block per frame load in the colorbar. This is to prevent colorbar stacking and improve efficiency for animations with large amounts of blocks.
- I fixed the **grouptags** parameter so that when called, animate creates separate animations for each loaded tag. It did not seem to be doing this before, so revert this if I messed it up.
- I created a **globalrange** helper function. This function allowed the **float** parameter to work for normal, grouptags, and multiblock applications with reduced redundancy. 

## GData

1. **_neighbors** attribute
This attribute is a dictionary which contains information on what blocks are next to (neighbor) a specific block. Each key in this dictionary represents a direction. The keys are tuples where the first value is an index which represents the dimension (0 or 1) and the second value is a boolean which represents the direction in that dimension (True=up/right and False=down/left). So the 4 keys in the _neighbors attribute are (0, True) for east, (0,False) for west, (1,True) for north, and (1,False) for south. The value at each key is the block object in that direction, if there is no block (an edge) then the value is None. The _neighbors attribute starts with None in every direction, but the subsequent set_neighbors method I will discuss is used to populate the attribute. The applications I have been testing this on do not need neighbors in more than 2 dimensions, but more may be added to the dictionary if needed.
2. **set_neighbors** method
This method uses the loaded block objects’ grid boundaries to identify which blocks are next to a specified block and what direction they are in, then it updates the information in that block’s _neighbors attribute. This method will most likely be redone when the multiblock topology files are finished and implemented since this method is not robust enough for applications that include multiple types of coordinate systems or blocks overlapping each other.

The addition of this attribute and method to GData is to support mutliblock lineouts. By recursively calling the set_neighbors method until the _neighbors value in a certain direction becomes None, one can find all the blocks which are included in a lineout and plot/stitch them together.

## Select
The way that select works normally has not changed. For multiblock cases, I added a **multiblock** parameter flag. My algorithm for multiblock lineouts is to first identify the bottom left point in order to find the minimum common edge in the z0 and z1 directions. Then it uses this information to find the starting block on the corresponding edge which starts the lineout. Then the _neighbors attribute and set_neighbors method are used to identify all the blocks in the lineout’s direction. The internal select function is run on all of these blocks and the resulting grids and value lists are stitched together and assigned to a new data object which is outputted. The previously loaded block objects are deactivated and the new data object is pushed into the data context. This means that when plotting the lineout the **multiblock** parameter is not needed.

### Example code:
`pgkyl <blocks in frame to load> sel  --z(#) <coordinate> --multiblock (or -mb) plot`
(#) is the dimension you want to perform the lineout in (0 or 1)

In order to support animation for these types of lineouts, I implemented a new structure into select similar to the one I used in animate. Before performing a lineout, each loaded block object is assigned its frame using the **set_frame** function and organized into a list of lists where each interior list contains the blocks associated with a frame. A multiblock lineout is then applied onto each frame and the new data objects are outputted. The default for **mutliblock** select is to assume that all loaded data objects are blocks belonging to the same frame, so to specify that there are multiple frames being loaded, a **multiframe** parameter is needed. This parameter is necessary when doing multiblock lineout animations or plotting multiple multiblock lineouts at the same time. Like the aforementioned single frame multiblock lineout plotting, the **multiblock** parameter is not needed since all outputs are singleblock 1D lineouts.

### Small fix:
I noticed that when calling the select, I could create and plot a lineout for a coordinate well outside of the data’s grid boundaries. So, in the select function in the data directory, I created a conditional which does not let it continue if the selected coordinate is outside of grid boundaries by a factor of 25% of the boundary range.

### Example code:
`pgkyl <blocks in all frames to load> sel --z# <coordinate> --multiblock (or -mb) --multiframe (or -mf) animate`

This link contains a slideshow with some examples of plots, animations, lineouts, and some test code: [link](https://docs.google.com/presentation/d/1J9aFJkWhbQFoaeml7SMoFKc5acHH3qssimBg1OsD31k/edit#slide=id.g2f41c82def6_0_0)


src/postgkyl/commands/plot.py
pcagas, line 1: Similarly to `animate` use just a single character for the short parameter names starting with `-`.

src/postgkyl/commands/select.py
```diff
@@ -29,19 +34,117 @@ def select(ctx, **kwargs):
...
-      data.add(out)
+  #multiblock case
+  if kwargs['multiblock']:
+    
+    #set ctx frames
+    frame_list = set_frame(ctx)
+    #creates list of lists with blocks per frame if multiframe parameter
+    #if not, then only one frame with all blocks
+    if kwargs['multiframe']:
+      dataList = []
```
pcagas, line 45: Please use `data_list`

src/postgkyl/utils/set_frame.py
pcagas, line 1: public function should be typed; i.e. all the inputs should have types and the function itself should have a return type. i think this should be:
`def set_frame(ctx : click.core.Context) -> list:`

In addition, there should be a docstring describing what the function does. I'm aware that not all public functions currently have it but it is being added and new functions should have it from the start.

src/postgkyl/data/select.py
pcagas, line 1: Double quotes please for consistency.

src/postgkyl/data/gdata.py
pcagas, line 1: Even though this is not needed for testing, it should not be hardcoded for 2D.

I personally don't like this dictionary keyed by tuple contract and especially the use of True/False for direction. I think it would be better to either split it into simple lists of lower and upper neighbors or, maybe even better, a list of `[lower_neighbor, upper neighbor]`.

pcagas:
@lucaa-g I did forget one thing in my review! You and @JonathanGorard need to be added to `AUTHORS.md` :-)

lucaa-g:
Hi @pcagas, thank you for the review! I have fixed the formatting issues (snake case, loop-ending comments, etc.), deleted the binary files, and added documentation to set_frame(). I have also changed the key/value makeup for the _neighbors attribute. Instead of using a tuple (dim: int, direction: boolean) as the key for the neighboring blocks, I have changed it so that the key is the dim (int) and the value is a list of the neighboring data objects in both directions [lower_neighbor, upper_neighbor]. Is this what you had in mind in your comment? I have also changed select so that it incorporates this different format. I have requested another review so please take a look at my changes and let me know if there are any fixes I should make. Thanks again.

src/postgkyl/data/gdata.py
pcagas: This is moving in the right direction!
A few comments on the changes:

1. A `dictionary` keyed with an integer is just a `list`, right? Why do you use `dictionary` instead of `list`? Are there any of the special `dictionary` features needed?
2. Instead of  `self._neighbors[dim] = [data, self._neighbors[dim][1]]` you can use  `self._neighbors[dim][0] = data`.
3. This is still hardcoded for 2D.
4. If I understand this correctly, it is looping through all the datasets for all the datasets and then comparing if the grid extends are the same, right? As a side note, comparing floats with `==` can lead to problems; using something like `numpy.isclose()` is often better. But more importantly, is this how we want to do this? @ammarhakim and @JonathanGorard, we already do have this information available, right? That might also naturally resolve issues with the dimensionality.

pcagas (changes requested):
@lucaa-g this is already much better, thank you!

I still have some concerns about how are the neighbors constructed; see my comment on the file. Maybe this needs to be discussed with @ammarhakim and @JonathanGorard?

There is also one more thing I completely forgot to mention last week, sorry about that :-( For any new feature like this we need to be adding tests. Please pick a reasonably sized file, copy it to `test_data` with some reasonable name, and add the tests themselves. I think we need at least a load test and possibly a plot test as well. If you are unsure how to do this, I can definitely help!

pcagas (approved):

PR #152, opened 2024-11-26
Testing

pcagas:
When Adios is not installed, the Adios tests are correctly skipped instead of failing; the same is true for FFmpeg.


ammarhakim:
So it is ready to merge? The tests are failing

pcagas:
should be all good now, @ammarhakim

PR #158, opened 2025-04-14
Gkyl partial load

pcagas:
This PR adds the capability of subselecting data _before_ loading the whole dataset into memory.

This now works for both type 1 and type 3 data. See the example of HLL Euler test data:

```
pgkyl hll-euler.gkyl pl -a
```
image

```
pgkyl hll-euler.gkyl --z0 '10:-5' -c '0' pl -a
```
image

```
pgkyl hll-euler.gkyl --z0 '10:-5' --z1 '25' -c '0' pl -a
```
image


manauref:
what's type 1 and type 3 data?

also, @pcagas , I assume once you load partial data you can perform the usual manipulations on the partial data (e.g. interp, ev, etc), right?

JunoRavin:
@manauref type 1 is single block data and type 3 is multi-block data. And yes, you can do all the usual manipulations to the data once it's loaded. I'm currently testing this functionality on some distribution function data I have that's 10GB pre-interpolation (and would be >100GB if I interpolated the whole field) and it's working just fine to interpolate, further down-select, integrate, etc.

PR #164, opened 2025-07-16
Subplot titles and labels

Maxwell-Rosen:
One of my long-standing frustrations with postgkyl has been its handling of subplots and the axis labels. I added an option to assign individual titles and axis labels to each subplot. Furthermore, to reduce clutter on the plots, I changed it only to label the outer axes, adding a super title and super axis label, so that the same information is not repeated multiple times.

Here is a sample using the following command:
` pgkyl tests/test_data/gk_wham-ion_integrated_moms.gkyl pl --title "integrated moments" --xlabel "x" --ylabel "y" --subplot-titles "a,b,c,d" --subplot-xlabels "t,y,u,i" --subplot-ylabels "q,w,e,r" --no-legend`
image

### DIFF
--- a/src/postgkyl/data/gkyl_reader.py
+++ b/src/postgkyl/data/gkyl_reader.py
@@ -146,29 +146,12 @@ class GkylReader(object):
       self.partial_idxs[6] = str(comp)
     #end
 
-  def _read_from_file(self, dtype, count, offset):
-    """Helper method to read from file with offset, compatible with NumPy 2.0+.
-    
-    Args:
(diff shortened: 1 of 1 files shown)


Maxwell-Rosen:
It looks like they started failing around #161, so these PRs are not the cause of the CI failing. I can investigate

Maxwell-Rosen:
I did break some unit tests. Some of them were checking for the ylabel, which I moved to be a supylabel for the whole figure. CI is working in both branches now

PR #165, opened 2025-07-17
Add new method to ev which simply scales the components of an array by a constant

Maxwell-Rosen:
This is a **very** needed feature for the BiMaxwellian moments output. The temperatures are in T/m, but usually we present them in electron volts. This allows us to do it in a simple command:

Example use
pgkyl data.gkyl ev 'f 2,3 2932384 scale_comp'

The first argument is the data structure, the second argument is the indices we scale (can be expressed as 2,3 or 2:4), the third argument is the constant to scale things by, and last we call the operation. 

I have tested and we can perform operations inside the scale coefficient, so one could do 'f 2.3 4 5 * scale_comp' and it works fine.

Example uses for the comp_scale function.

Base case: 
`pgkyl gk_wham_1x2v_p1-ion_BiMaxwellianMoments_0.gkyl interp pl`
image

Multiply components 2 and 3 by 100000000
`pgkyl gk_wham_1x2v_p1-ion_BiMaxwellianMoments_0.gkyl interp ev 'f 2:4 100000000 scale_comp' pl`
image

Only scale components 0 and 3 by 100000000:
`pgkyl gk_wham_1x2v_p1-ion_BiMaxwellianMoments_0.gkyl interp ev 'f 0,3 100000000 scale_comp' pl`
image

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -157,7 +157,7 @@ class TestCommands:
     fn = tmp_path / "test_anim.mp4"
     self.ctx.invoke(cmd.animate, show=False, saveas=fn)
     fig = plt.gcf()
-    label = fig.get_axes()[0].get_ylabel()
+    label = fig.figure.get_supylabel()
     self.ctx.obj['data'].clean()
     self.ctx.obj["in_data_strings_loaded"] = 0
     plt.close("all")


Maxwell-Rosen:
I also want to be able to scale the axes of the plots, so I added a new method. Here are some uses and examples of it working.
```
pgkyl gk_wham_1x2v_p1-ion_0.gkyl interp sel --z0 0.0 pl &         
pgkyl gk_wham_1x2v_p1-ion_0.gkyl interp sel --z0 0.0 ev 'f 1 100 scale_zi_axis' pl &
pgkyl gk_wham_1x2v_p1-ion_0.gkyl interp sel --z0 0.0 ev 'f 2 100 scale_zi_axis' pl &
```
image

manauref:
> I also want to be able to scale the axes of the plots, so I added a new method. Here are some uses and examples of it working. It's helpful for scaling the axis for things like U/c_s or v/v_th.

Youn can already to do this with plot `--xscale` and `--yscale`

Maxwell-Rosen:
There is an advantage to doing it with ev rather than plot. With ev, you can resave stretched and compressed data to new formats. I tried to do this a few months ago with some older simulations and there wasn't a method to stretch and compress the x/y axes then resave it. It's good to support multiple ways of solving the same problem

PR #166, opened 2025-08-26
Fixing grid deletion issue. 

jtenbarg:
Previously attempted to use np.delete on a list rather than a numpy array. Switched to using pop() instead.

### DIFF
--- a/src/postgkyl/tools/fft.py
+++ b/src/postgkyl/tools/fft.py
@@ -40,11 +40,11 @@ def fft(data: GData, psd: bool = False, iso: bool = False,
     # end
   # end
   if idx:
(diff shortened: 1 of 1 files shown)


src/postgkyl/tools/fft.py
```diff
@@ -40,11 +40,11 @@ def fft(data: GData, psd: bool = False, iso: bool = False,
     # end
   # end
   if idx:
-    grid = np.delete(grid, idx)
+    #grid = np.delete(grid, idx)
```
JunoRavin: @jtenbarg instead of commenting out previous lines, we should just delete this if this fix is robust here so that way we don't confuse ourselves later.

PR #174, opened 2025-12-18
Gridplot tool

mp5650:
Added gridplot tool to postgkeyll. Only file I changed was adding grid plot.py to src/postgkyl/commands  and then declaring it in src/postgkyl/commands/__init__.py and src/postgkyl/pgkyl.py

 Usage: pgkyl gridplot <name>
  
 Expects a directory like '<name>_geometry/' to exist., the files '<name>_bX-nodes.gkyl', and '<name>_settings.gkyl' must exist in that directory. 
 Directory and relevant files can be generated using tokagridgen tool.


manauref:
closing in preparation for pgkyl refactor. We'll re-evaluate the features in this command after that.

PR #190, opened 2026-05-06
Add apar energy rate of change to the balance

Antoinehoff:
see description.

We also add `plt.tight_layout()` commands before showing or saving to avoid cropping.

### DIFF
--- a/src/postgkyl/commands/gk_energy_balance.py
+++ b/src/postgkyl/commands/gk_energy_balance.py
@@ -520,7 +520,8 @@ def gk_energy_balance(ctx, **kwargs):
   ax1a.set_title(title_string,fontsize=title_font_size)
   ax1a.set_xlim( time_fdot[0], time_fdot[-1] )
   set_tick_font_size(ax1a,tick_font_size)
-
+  
+  plt.tight_layout()
   if kwargs["saveas"]:
     plt.savefig(kwargs["saveas"])
   else:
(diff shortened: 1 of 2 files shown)


manauref:
great!
can you add the requirement on this file to the command description (which appears when using `-h`) please?

Antoinehoff:
> great! can you add the requirement on this file to the command description (which appears when using `-h`) please?

Done

src/postgkyl/commands/gk_energy_balance.py
```diff
@@ -369,9 +404,18 @@ def absy_disabled(data_in):
...
+        combined_field_dot = field_dot + apar_dot
+        gdat_field_dot.push(time_field_dot, combined_field_dot)
+      else:
+        gdat_field_dot.push(time_field_dot, field_dot)
       data.add(gdat_field_dot)
 
+    if has_apar_dot and not has_field_dot:
+      gdat_apar_dot.push(time_apar_dot, apar_dot)
+      data.add(gdat_apar_dot)
+
```
manauref: @Antoinehoff why combine the ES and EM contributions? why don't we plot separate lines for these two?

Antoinehoff:
Here is an example of a plot for the alfven wave regression test 3x2v with nperiod=1
Screenshot 2026-06-17 at 2 11 20 PM

manauref (approved):

PR #193, opened 2026-05-15
3D plotting methods to command line postgkeyll

Maxwell-Rosen:
# Feature

## Overview

This PR adds two new methods to the command-line/script pgkyl, utilizing the 3D plotting libraries plotly and pyvista. I first implemented the Plotly version, but was dissatisfied with the rendering and performance. I implemented a PyVSITA backend, but there are some tradeoffs, so I decided that both are valuable. While plotly renders in browsers, making saving and sharing simple, pyvista renders using Python's VTK library, so it's in dedicated windows, and it has much better performance.

From the command line, the new commands are `plotly / ply`, `plotly-animate / ply-anim`, and `pyvista / pv`. I introduced some aliases for these commands, which are not just the shortened versions, so that we don't have to spell everything out, and it remains in a condensed but meaningful notation. This is important so that we can retain our functionality of `pl` specifying a matplotlib plot.

In doing this, I had to refactor the plotting scripts, as they have some shared functionality. I based this on the utilities directory, where there are some shared functions that are private to the utilities modules.

I refactored the write method and added support to save *.vtk objects, which were necessary to import postgkeyll data into Paraview, which I used for a VR visualization. *.vtk is a universal file format for this kind of structured data.

A major limitation I’ve had with PyVista is that the axes normalization can be very strange. I had numerous visual bugs and errors because PyVista interprets all input axes as "meters," but we often deal with incredibly stretched grids, and it doesn't have built-in axis rescaling options for scientific data. I have a re-labeling procedure, but it's pasted on, and I noticed several issues with this in development, which have been addressed by now. On the other hand, Plotly has a very nice built-in axes construction. This even led to a PR into pyvista itself because I found an issue with their API (https://github.com/pyvista/pyvista/pull/8590).

One of the most useful things about the plotly backend is the mountain height maps for 2D data, so that we aren't solely relying on color to measure value, but can see the "height" z of the data in 3D.

I've been testing on several forms of data and can showcase some results below.

When constructing this PR, my work with Plotly was heavily using LLMs, and I think that's pretty evident in the work. After some time, I ran out of tokens, so I had to resort to the barbaric methods of coding myself. The pyvista module was written by me, which is why it looks so nice. On the other hand, plotly can do other things that pyvista cannot do (height maps / scatter plots) and has very different constraints. I tried to make the plotly version as readable and maintainable as possible. This code is very maintainable if you use LLMs.

### Extensions

Animate works with plotly, but I didn't get it working in pyvista. This is left to future work. It could be a command like `pyvista-animate`.

### Samples

`pgkyl gk-distf -n gk_lorentzian_mirror -s ion -f 75 -v ev 'f abs'  plotly --mppa 32`
image

# Distribution function in 3d. Log scale, with colorbar limits
pgkyl gk-distf -v -n gk_lorentzian_mirror -s ion -f 75 ev 'f abs' plotly --mppa 32 --surface-count 32 --logc --cmin 1e-15 --xlabel "Z, m" --ylabel "v_{\parallel}, m/s" --zlabel "\mu, J/T" --clabel "|f|" -o 1.0 --saveas distf.html 
image

# Supports light mode as well. Can change colormaps. Invert using --invert-cmap
pgkyl gk-distf -v -n gk_lorentzian_mirror -s ion -f 75 ev 'f abs' plotly --mppa 32 --cmap "viridis" --invert-cmap --surface-count 32 --logc --cmin 1e-15 --xlabel "Z, m" --ylabel "v_{\parallel}, m/s" --zlabel "\mu, J/T" --clabel "|f|" -o 1.0 --saveas distf-light.html --background light
image

# Linear colorbar reveals different features
pgkyl gk-distf -v -n gk_lorentzian_mirror -s ion -f 75 ev 'f abs' plotly --mppa 32 --surface-count 32 --xlabel "Z, m" --ylabel "v_{\parallel}, m/s" --zlabel "\mu, J/T" --clabel "|f|" -o 1.0 --saveas distf-linear.html
image

# This works for any kind of data!
# Here is the density from a TCV simulation
pgkyl tcv-data/rt_gk_tcv_nt_iwl_adapt_src_3x2v_p1-ion_M0_64.gkyl interp plotly --mppa 32 --surface-count 32 --saveas "tcv-log.html" --logc --cmin 1e16 --clabel "Density, m^{-3}"
image

# Support scatter plot for mapc2p data
pgkyl --c2p tcv-new-data/rt_gk_tcv_nt_iwl_adapt_src_3x2v_p1-mapc2p.gkyl tcv-new-data/rt_gk_tcv_nt_iwl_adapt_src_3x2v_p1-ion_M0_0.gkyl interp ply -s --background light --saveas "tcv-cartesian-scatter.html" --aspect data --scatter-opacity-range 0.05,1.0
image

# One of the coolest things is plotting 2D data as a height plot, like mountains
# Mirror distribution function at the midplane
pgkyl gk-distf -v -n gk_lorentzian_mirror -s ion -f 75 sel --z0 0.0 plotly --xlabel  "v_{\parallel}, m/s" --ylabel "\mu, J/T" --zlabel "f" --clabel "f" --saveas distf-midplane-surface.html
image

# Pyvista examples
# Pyvista can load more data. It's better optimized, as it doesn't run in a web browser.
# No downsampling
pgkyl gk-distf -n gk_lorentzian_mirror -s ion -f 75 -v ev "f abs" pyvista --logc --contour-levels 20 --cmin 1e-15
image

# Neet widgets to slice data. Can even do off-axis!
pgkyl gk-distf -n gk_lorentzian_mirror -s ion -f 75 -v ev "f abs" pyvista --logc --contour-levels 20 --cmin 1e-15 --mesh-clip-plane --opacity 1
image

# Can do slices without contours
pgkyl gk-distf -n gk_lorentzian_mirror -s ion -f 75 -v ev "f abs" pyvista --logc --cmin 1e-15 --mesh-slice-plane --opacity 1 --no-contour
image


src/postgkyl/data/write.py
Antoinehoff, line 1: What's the goal of this file? command/write.py is already here, it seems that it is some duplication?

src/postgkyl/utils/axis_and_grid_prep.py
Antoinehoff, line 1: is this of general utility? I think it could go to utils too.

src/postgkyl/utils/downsample.py
Antoinehoff, line 1: this should also go to utils

src/postgkyl/data/write.py
Maxwell-Rosen: Hmmm, yes, I did bundle this together. On main, this method is bundled inside the GData object. Instead, I refactored this outside the object so it’s clear that the command has an associated file it's calling. I added a few new methods to the write file, including saving 3D data to import into Paraview.

src/postgkyl/utils/axis_and_grid_prep.py
Maxwell-Rosen: Move to utils. This is a shared function among the 3 plotting scripts. It's only used for plotting, but I could see it being useful to script users who want a clean command for organizing the grid and axis.

src/postgkyl/utils/downsample.py
Maxwell-Rosen: I agree. Moved.

manauref (changes requested):
please simplify this work. It's unnecessarily complicated and unmaintainable (without LLMs).

Please remove one of the options, plotly or pyvista. Let's only use 1 in postgkyl.

Maxwell-Rosen:
I strongly disagree with this feedback that these modules should cut features to improve readability. There is a very good reason the plotly script is much longer than the pyvista one: it has many more features, and they are important. Plotly and pyvista serve different purposes, and including both is valuable.

Pyvista
- Only 3D volume / contour plotting + wigits (native, simple flags)

Plotly
- 3D volume /contour plots
- 3D scatter plots
- 2D surface height plots
- Interactive widgets inside the application to control spinning
- Supports animations

5x the features for 5x the size is not bloat -- it's linearly proportional. I will do my best to refactor the plotly script to be more readable, but I've already refactored this file. Further refactors would just be creating more helper files rather than removing code.

ammarhakim:
I would like this PR to be accepted. @manauref Please remove the block and we will take the simplification  to a later set of work. Having two libraries is fine. Pgkyl already depends on too many things and one more is not going to make any difference.

ammarhakim (approved):
This is ready to merge.

Maxwell-Rosen:
Merged into #229

PR #195, opened 2026-05-19
Add command line fitting tools

Maxwell-Rosen:
This is nearly 100% Claude, with my guidance. I approve these changes.

I wanted to fit a line through some data in command line pgkyl, so I created a new command. This wraps scipy.optimize and their curve fitting package. I aimed for syntax like 
`pgkyl data.gkyl fit linear`
which would output the linear coefficients of the data, with R^2 and error in the fit. I implemented a few other types of curves, such as quadratic, exponential, sigmoid, plateauing exponential. This command loads data to the stack so you can
`pgkyl data.gkyl fit lin pl`
and it plots the fit on top of the data. It works for multi-component data.
Guesses are constructed for each function based on the type of equation. Custom functions can be specified as
`pgkyl data.gkyl fit 'a x * b +'`
 in reverse polish notation (ugly). This makes it very extendable, but you should include a good guess for the parameters when using custom functions, as our data tends to have very large ranges and all variables have initial guesses of 1.0.
Due to the refactor of plotting methods and output, this is a queued merge behind 3d plotting.

Note that there is structure to this work
pgkyl/commands has the click wrapper and simply wraps the click
pgkyl/tools has the real computation and the operation
This structure makes it simple to interface with script python and we should incorporate this structure into the other modules as well. It would be good to aim for a simple click wrapper for command line, which wraps the script-based functions. Some parts are like this, but it's not consistent.

Example
```
pgkyl gk_lorentzian_mirror-ion_integrated_moms.gkyl fit exp pl -f0
default
  Component 0:
Exp plateau: y = (-1.586212e+18 ± 1.89e+15)*exp((-4.128902e+00 ± 1.50e-02)*x) + (1.610502e+18 ± 2.47e+15)    R² = 0.997291
  Component 1:
Exp plateau: y = (3.471025e+20 ± 5.29e+23)*exp((-1.161688e-02 ± 1.86e+01)*x) + (-3.470651e+20 ± 5.29e+23)    R² = 0.000233
  Component 2:
Exp plateau: y = (-1.652957e+30 ± 1.28e+28)*exp((-2.001301e+00 ± 2.61e-02)*x) + (1.773358e+30 ± 1.39e+28)    R² = 0.991441
  Component 3:
Exp plateau: y = (-4.633854e+30 ± 5.83e+27)*exp((-2.301744e+00 ± 5.32e-03)*x) + (4.570428e+30 ± 6.50e+27)    R² = 0.999643
```

image


Maxwell-Rosen:
Merged into #229

PR #202, opened 2026-06-15
[PR] Switch plt.subplots to fig.subplots in plot script

Antoinehoff:
Latest version of Matplotlib does not like when we pick a figure that is already defined with plt.suplots. We just call fig.subplots instead.


manauref:
This seems safe. I tried it in my conda build and the plot works as before.

PR #205, opened 2026-06-22
add a new command to work on discontinuous representation of the data

Antoinehoff:
We implement here a function to plot the discontinuous representation of any field. It must be in 1D for now. If you have ideas to do some discontinuous oriented planes in 2D I'm interested!

here is an example of what you can do
```
pgkyl /pscratch/sd/a/ah1032/gkeyll_em/tcv_pt_0.5MW_coarse/smoothapar/wk_restart/rt_gk_tcv_nt_iwl_adapt_src_3x2v_p1-ion_M0_237.gkyl dg select --z1 0.0 --z2 0.0 plot --figure 0 --no-legend --xlabel '$x$ [m]' --ylabel '$n_i$ OBMP' --clabel '$t$ [$\mu$s]'
```

image


manauref:
@Antoinehoff so I think this should take a data from the stack (you do this already), and in the case of multidimensional datasets, I think it should take the output of the `select` command.

However, I know select is not quite ready to do that is needed here. Specifially, select would have to:
1. Evaluate the DG field at a specific floating point coordinate.
2. Project onto the corresponding lower-dimensional basis.

So, I propose that we upgrade `select`so it can do these things, then we can simply pass a 1D field to gk-dg, and this new command wouldn't have to do any selection (it's best to keep operations to the tools we already have when possible).

Antoinehoff:
> @Antoinehoff so I think this should take a data from the stack (you do this already), and in the case of multidimensional datasets, I think it should take the output of the `select` command.
> 
> However, I know select is not quite ready to do that is needed here. Specifially, select would have to:
> 
> 1. Evaluate the DG field at a specific floating point coordinate.
> 2. Project onto the corresponding lower-dimensional basis.
> 
> So, I propose that we upgrade `select`so it can do these things, then we can simply pass a 1D field to gk-dg, and this new command wouldn't have to do any selection (it's best to keep operations to the tools we already have when possible).

agreed. With an upgraded `select` we could plot the value at the surface, which can be very useful to check what is happening at boundaries.

Antoinehoff:
A good point raised by @JunoRavin, we should not call it `gk-dg`, but something more generic since it can be used beyond the GK world. One limitation is that it is not available if `p` if not `1`.

manauref:
don't just call it dg, that's not specific enough (we have lots of dg things).
maybe plot-dg-poly ? or plot-pw-poly?
of if this command is not going to do the plot, call it dg-pw-poly or something

Antoinehoff:
> 

yeah I agree, though it does not plot. Let's see how it goes once we did the work with the selection thing.

Antoinehoff:
waiting for other updates

Issue #199, opened 2026-05-29
[DR] human readable postgkeyll for scripts


Maxwell-Rosen:
I was inspired by a YouTube video about readable code.
https://m.youtube.com/watch?v=SJocPm2E8eQ
Part of my most recent frustrations come from the disconnect from script pgkyl and command line pgkyl have a very different interface. This leads to people aligning with either one or the other. It would be great to have a more command-line like interface we could use in pgkyl, similar to this. In contrast to my earlier DR's, I'm suggesting thinking about the script mode pgkyl from a top down approach. It will benefit us to think about the final scripts we want to write, rather than an API-first approach.  I know this is a standard goal, ideal, and objective for our projects, but I don't think it's been applied well to script pgkyl. 

In a Python script, we could plot some data like

```python
Import postgkeyll as pg

pg.load('filename').select(z0=0.0).plot()
```

This could also write the common data object to avoid using activate and tagging.

```python
Import postgkeyll as pg

data1 = pg.load('filename').select(z0=0.0)
data2 = pg.load('file2').evaluate('f 5 +').select(z0=0.0)
plot(data1.and(data2))
print(data1)
```

The goal is to construct a "language" for how to write pgkyl scripts, similar to the YouTube video describes. In practice, this could be an object oriented wrapper for our underlying functionality. I think these examples even point out that things seem out of order when reading command like pgkyl. 

We can define a dunder method for these data classes to perform regular Python operations on these objects (print, add, multiply, divide). This would make manipulation more intuitive.

I would like to get the perspective of Claude for this interface. It might have intelligent comments on the design.

Similarly, we could load entire simulations like

```python
Import postgkeyll as pg

pg.load.simulation('simulationName', GK, 1X2V)
```

Maxwell-Rosen:
Here is what Claude came up with for a plan of structuring the API. It seems interesting.

[READABLE_API.md.rtf](https://github.com/user-attachments/files/29266719/READABLE_API.md.rtf)

ammarhakim:
This is complete slop. Not interested in this. I am closing this so-called DR. God knows what it is about and I am not interested in this at all.

Maxwell-Rosen:
[...]

PR #206, opened 2026-06-22
👏 CLAP interface

Antoinehoff:
This PR implements an order zero API help us use postgkyl in Python.

### Criticism
This API is the order zero of an API in the sense that it wraps on top of the CLI which is not very elegant. Instead of doing
```
API->CLI->data/processing/plots
```
we should do
```
CLI->API->data/processing/plots
```
but this way more tedious and asks a serious refactoring of the library. No one is paid for that right now so until unconditional revenues I think we can keep this API and just name it the Command Line Application Programming Interface, or **CLAP👏 interface**.

### Main design choices
**We don't want to manually implement an interface of every clickable command**.
To avoid that, we have an API generator `src/postgkyl/api/_api_gen.py` that will scan through the available commands and create the API automatically in a new file `src/postgkyl/api/api.py`. We also add a specific README for the API that recalls how to use and update it if necessary.

I've tested the robustness of this API in many ways, in particular when implementing a new command, and it does not require any change, we just need to run `python src/postgkyl/api/_api_gen.py` and this will update the API.

To have a state that will stack the data over multiple, separated calls, we have a new `src/postgkyl/api/scripting.py` file that defines the `_Session` class. This class allows us to build up the command line by line. This file is not automatically generated and can be used to have useful tools like `_Session.get_framelist` that will get you the list of frame for a given field or `_Session.print_cmd` that will print the equivalent CLI code for the current state of the API session.

### A few examples
We can now write more human understandable python code that will make full use of the `postgkyl` commands.

For example, this command
```
pgkyl /Users/ahoffman/personal_gkyl_scripts/sim_data/3x2v/gk_tcv_adapt_src/rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_0.gkyl interpolate select --z2 31 plot --logz --xlabel '$R-R_{lcfs}$ [m]' --ylabel '$y$ [m]' --clabel '$n_i$ [m$^{-3}$]' --colormap twilight
```

Can be called as
```python
from postgkyl.api import PgkylSession
pg = PgkylSession()
pg.load(
    "/Users/ahoffman/personal_gkyl_scripts/sim_data/3x2v/gk_tcv_adapt_src/rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_0.gkyl"
)
pg.interpolate()
pg.select(z2=31)
pg.plot(
    xlabel=r'$R-R_{lcfs}$ [m]',
    ylabel=r'$y$ [m]',
    clabel=r'$n_i$ [m$^{-3}$]',
    logz=True,
    cmap="twilight",
)
```

both produce

image

We also add the possibility to provide a colormap when passing many 1D plots to create a profile time evolution plot, i.e.
```python
from postgkyl.api import PgkylSession
pg = PgkylSession()
frame_list = pg.get_framelist(
    name="apar",
    simprefix="rt_gk_em_alfven_1x2v_p1",
    path="/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/")
frame_list_str = [str(i) for i in frame_list]
# sum the frame str like 1,2,3,4
frame_list_str = ",".join(frame_list_str)
for frame in frame_list:
    pg.load(
        f"/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_{frame}.gkyl"
    )
pg.interpolate()
pg.plot(figure=0,
        cmap="viridis",
        no_legend=True,
        xlabel='z',
        ylabel='Apar',
        clabel='frame',)
```
produces 

image

Using `pg.print_cmd()` yields:
```
pgkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_0.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_1.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_2.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_3.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_4.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_5.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_6.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_7.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_8.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_9.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_10.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_11.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_12.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_13.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_14.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_15.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_16.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_17.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_18.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_19.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_20.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_21.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_22.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_23.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_24.gkyl /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_alfven_1x2v.o/rt_gk_em_alfven_1x2v_p1-apar_25.gkyl interpolate plot --figure 0 --no-legend --xlabel z --ylabel Apar --clabel frame --colormap viridis
```
 I dare @manauref to come up with these kinds of command manually.

### Additional notes
There are a a few additional features present in other PR in this branch because I want to make sure the API generator is robust to new incoming commands.


Maxwell-Rosen:
Can you comment on how to perform python-native operations on the data? For instance, it would be very elegant if we could do something like 
```python
a = pg.load('file-a.gkyl')
b = pg.load('file-b.gkyl')
c = a + b
pg.plot(c)
```
Such a syntax makes data manipulation much more readable, customizable, and straightforward. This could be accomplished by using the structure laid out in #199. 

Additionally, I don’t think this is the correct approach to layer the click statements. It would be cleaner if click were a thin wrapping layer for an underlying script-oriented interface. Ideally, the click wrappers should be thin and compact. This makes development and the addition of new features much simpler. It adds unnecessary layers if we wrap it in the order `python API -> commands -> python functions`. Instead, wrapping in the order’ commands -> python API -> python functions' would create a simpler development path for the CLI.

Additionally, I am not aware of many Python packages that require an initialization command. I would think that we should aim for a more standard syntax with other packages in the language, such as numpy or scipy.

Furthermore, CI should not be failing. It's passing on main AFAIK. Please look into this.

Antoinehoff:
> Can you comment on how to perform python-native operations on the data? For instance, it would be very elegant if we could do something like
> 
> ```python
> a = pg.load('file-a.gkyl')
> b = pg.load('file-b.gkyl')
> c = a + b
> pg.plot(c)
> ```
> 
> Such a syntax makes data manipulation much more readable, customizable, and straightforward. This could be accomplished by using the structure laid out in #199.

I don't think you got the meaning of this PR. It's an interface to the CLI nothing else. You can do any operation you can do with CLI.
Then for your idea here's an exercise for you: how would you do `c=a/b`?

> 
> Additionally, I don’t think this is the correct approach to layer the click statements. It would be cleaner if click were a thin wrapping layer for an underlying script-oriented interface. Ideally, the click wrappers should be thin and compact. This makes development and the addition of new features much simpler. It adds unnecessary layers if we wrap it in the order `python API -> commands -> python functions`. Instead, wrapping in the order’ commands -> python API -> python functions' would create a simpler development path for the CLI.
> 

This PR is a small wrapper to allow to script easily. What you ask is a deep rewrite of the library.

> Additionally, I am not aware of many Python packages that require an initialization command. I would think that we should aim for a more standard syntax with other packages in the language, such as numpy or scipy.
> 
I don't see why you compare this to numpy or scipy.

> Furthermore, CI should not be failing. It's passing on main AFAIK. Please look into this.

Yes.

Maxwell-Rosen:
My takeaway is that this PR makes the scripting very nice. It's much nicer than the scripting we have. However, I think the structure and layering are backward, and this may punt a deeper refactor into the future.

The result will be similar to how pgkyl has developed, in that once the CLI cannot do what you're trying to do, you revert to the ugly scripting commands, which are difficult to read.

Yes, division is  `c = a/b`, which calls the division method, which can call weak division. Or it can be pointwise, however you want it. It's a bit confusing to divide interpolated point-wise data with DG data, but we can throw an error for that.

JunoRavin:
Just to chime in here, I do not view this PR as precluding the layering contract of commands -> python API -> python functions

This is a proposal to add another means of interfacing with the command line interface in script mode, which may be convenient in Jupyter or Marimo notebooks. 

All click-based commands can still be thin wrappers around functions that could be called in script mode (and indeed, this wrapping will only get more sophisticated as more of the Gkeyll core library is wrapped in postgkyl). 

Ultimately this question about about how something read in at the Python level can be manipulated in Python will be completely obviated by postgkyl being itself a wrapper around Gkeyll functionality up until the final point in which a user decides they're going to make a plot (and with @Antoinehoff's recent DG command, we may yet have further sophistication to visualize a wide variety of data in its raw DG form should we choose). 

That doesn't mean we should be supporting all of the object-oriented operator overload stuff Python supports. It means that our Python workflow should be operating on Gkeyll data in an identical fashion to if we were doing the analysis directly within Gkeyll (so we won't be making a + b make sense, we will be making a wrapper around array_ops and exposing the array_ops routines). Thus, large discussions about what the correct "Python-ification" of the library looks like are moot. We're not a Python library. We're a wrapper around Gkeyll. Full-stop. And we can utilize that wrapping however we choose (whether through scripting, CLI, or things like what @Antoinehoff has built here).

Maxwell-Rosen:
I understand this. We are wrapping C gkeyll with Python's mature plotting utilities. I am concerned about this approach because it is as limited as the CLI. Could you show me how you would load multiple datasets in a clean way, then plot without the CLI using your own custom plotting script? The issues with the CLI have been the convoluted data-loading and manipulation structure, for instance, recalling how we used to perform the correct scaling of the Jacobian for the distribution function.

```pgkyl --c2p-vel gk_lorentzian_mirror-elc_mapc2p_vel.gkyl gk_lorentzian_mirror-elc_0.gkyl -t jf gk_lorentzian_mirror-elc_jacobvel.gkyl -t jac ev -t df 'jf jac /' activ -t df interp -b gkhyb -p1 sel --z0 2.5 pl```

Would we use the same tagging and activation using this style? I find this unsightly and convoluted. One way I see to overcome tagging and activating is to use variables in a script.

Another concern is highlighted by another question. How would you divide a phase space array by a configuration space array? I am not sure this is possible in the CLI, which is why I occasionally have to use the underlying package.

I also hesitate to call this an API. It’s a wrapping of the CLI in a script interface, not a flexible programmable interface with Gkeyll itself.

**Comment by JunoRavin** [comment]
We're not going to get into arguments about what is unsightly in Python. The reality is that Python as a language is already excessively verbose and supports doing the same thing 12 different ways across multiple packages. 

We are going to allow users to engage with the tech stack in a number of ways. Script mode, CLI, and various home-baked APIs in script mode which further wrap postgkyl functionality. Indeed, across the whole project I know of at least a half-dozen bespoke wrappers around postgkyl itself already. The advantage of what @Antoinehoff is proposing is that you start to enforce some minor discipline in people who just want to wrap postgkyl themselves (by providing an efficient representation of the entire command stack that they can utilize however they'd like in their existing workflow). 

Will this be the way you specifically engage with postgkyl? Perhaps, but perhaps not. Again, this particular proposal is merely another wrapping of postgkyl. It does not preclude any larger refactors of reworked layering for script-mode, how to make the command line interface more robust, and our larger efforts to move post-processing operations down to the Gkeyll layer. Someone adding a mechanism to further wrap postgkyl in a way they find convenient, that is auto-generated, and is sufficiently modular with respect to larger postgkyl ambitions should not be dismissed merely because it doesn't conform to your taste in Python aesthetics, especially when again, your critiques of it are not relevant to the point of the design.

**Comment by Antoinehoff** [comment]
> I understand this. We are wrapping C gkeyll with Python's mature plotting utilities. I am concerned about this approach because it is as limited as the CLI. Could you show me how you would load multiple datasets in a clean way, then plot without the CLI using your own custom plotting script? The issues with the CLI have been the convoluted data-loading and manipulation structure, for instance, recalling how we used to perform the correct scaling of the Jacobian for the distribution function.
> 
> `pgkyl --c2p-vel gk_lorentzian_mirror-elc_mapc2p_vel.gkyl gk_lorentzian_mirror-elc_0.gkyl -t jf gk_lorentzian_mirror-elc_jacobvel.gkyl -t jac ev -t df 'jf jac /' activ -t df interp -b gkhyb -p1 sel --z0 2.5 pl`
> 

It's good that you provide me example of usage so that it stress tests the API. The API is really meant to build these kind of long command lines using a script, nothing else. I have to update the branch because I realized with this example that the options cannot be put wherever we want. E.g. `--c2p-vel` needs to be before the data, that I could understand for multi data loading I guess, and also `ev` needs a tag before the chain, which is less intuitive to me. Anyway, I'll come back on that soon.

**Comment by Antoinehoff** [comment]
Ok I wasn't aware of the CLI giving global arguments just after the `pgkyl` call that can be then accessed by the entire command stack. We can now pass these global arguments. I've checked and I can reproduce the example you provided Max, this is the equivalent python code:
```python
from postgkyl.api import PgkylSession
simdir = "/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_helical_1x2v.o/"
simprefix = "rt_gk_helical_zpar_adapt_src_1x2v_p1" 
fieldname = "elc_0"
pg = PgkylSession()
pg.set_globals(c2p_vel=f"{simdir}/wk_sgbc/{simprefix}-elc_mapc2p_vel.gkyl")
pg.load(
    f"{simdir}/wk_sgbc/{simprefix}-{fieldname}.gkyl",
    tag="jf"
)
pg.load(
    f"{simdir}/wk_sgbc/{simprefix}-elc_jacobvel.gkyl",
    tag="jac"
)
pg.ev('jf jac /', tag="df")
pg.activate(tag = "df")
pg.interpolate(basis_type="gkhyb", poly_order=1)
pg.select(z0=0.0)
pg.plot(xkcd=True, showgrid=False)
```
image

_I've added xkcd because we are all getting too serious in here_

I think this code, as well as the command line, will be simplified soon with the `gk-load` command in #209.

Again, thanks Max for stress testing, if you or other people want to check the interaction between this API and their favorite command lines please do!

Maxwell-Rosen:
A separate point to make, which I'm a bit confused about, is that we are now intending to implement a new foreign function interface with C Gkeyll, so we will have Lua and Python. Ammar has long protested an FFI with Python, in that he wants our input files in Lua. It's a decision he made that he no longer wants to argue about and is not up for debate. If we interface C Gkeyll with Python, will this be a full FFI? It doesn't sound like we will wrap the app system; rather, we'll just wrap minus or core. To keep with our conventions, shouldn't we do this in Lua? Otherwise, it would make much more sense to wrap Gkeyll entirely in Python so that we can handle inputs and postprocessing in the same script.

Maxwell-Rosen:
Another way to gain some flexibility is to be able to output the grid and values to variables, then reset the stack. This way, we can perform operations on the data or create unique plots, using the script wrapper to load data and perform simple manipulations. We could also add the ability to reload the data in script mode if that would be beneficial to the application. One example of this could be taking the distribution script above and dividing by the configuration space Jacobian. One could accomplish this by loading the configuration space data separately, outputting data / grids from both data sets, performing the division (by value at interpolated nodes), then re-loading this into pgkyl. Does this make sense? I think it would enhance the flexibility of this wrapper.

Another useful example of this would be plotting a slice of a 2D simulation on top of a 1D simulation. We would need to unload and reload data sets to do this, and it's not simple in the current script mode.

Maxwell-Rosen:
I love the rename to CLAP

PR #207, opened 2026-06-22
Parallel animate

Antoinehoff:
This implements the possibility to generate movies with a parallelization of the frame generation.

One can now run `animate` with `--nproc` command which will generate the animation using a number of processes bigger than 1. If `--nproc` is not specified, the code do not change.

We now ensure that the movie file given is a `.gif`, I don't know how to generate `.mp4` robustly and I think the user can use a gif 2 mp4 converter afterwards if needed.

tcv_nt_ion_M0

The animation above was generated on perlmutter login node. It took 30sec in sequential agains 12 with 4 processes, here is the command line 
```
pgkyl /pscratch/sd/a/ah1032/gkeyll_em/tcv_pt_0.5MW_coarse/smoothapar/wk_restart/rt_gk_tcv_nt_iwl_adapt_src_3x2v_p1-ion_M0_[0-9]*.gkyl gk-dg select --z1 0.0 --z2 0.0 -ylim 0,7e19 --saveas tcv_nt_ion_M0.gif --saveframes tmp/ --nproc 4
```

### note
For a very basic movie in 1D, We can easily divide the time to generate the animation by 2. Then the time start to be dominated by the postprocessing like `interp` `sel` etc.  
For heavier post processing task like `gk-rz` for multiple frames, one could consider having another layer in the API that would run the entire command in parallel to generate frames and then compiling the frames.

### DIFF
--- a/src/postgkyl/commands/animate.py
+++ b/src/postgkyl/commands/animate.py
@@ -40,15 +40,22 @@ def _save_frames(data_list, num_frames, prefix, kwargs, figsize, fig=None):
 # end
 
 
-def _compile_movie(frame_files, output_file, duration):
-  """Compile a list of PNG frame files into an animated GIF (or other PIL-supported format)."""
-  images = [Image.open(f) for f in frame_files]
-  print(f"Creating movie {output_file}...")
-  images[0].save(
-      output_file, save_all=True, append_images=images[1:],
-      duration=duration, loop=0, optimize=False,
-  )
-  print(f"Movie {output_file} created.")
+def _compile_movie(frame_files, output_file, fps, duration):
+  """Compile PNG frames into an animation."""
+  ext = os.path.splitext(output_file)[1].lower()
+  fps_val = fps if fps else 10
+  print(f"Creating {output_file}...")
+  if ext in (".gif", ".webp", ".apng"):
+    images = [Image.open(f) for f in frame_files]
+    images[0].save(
+        output_file, save_all=True, append_images=images[1:],
+        duration=duration, loop=0, optimize=False,
+    )
+  else:
+    # We do not support other format like .mp4 or .avi.
+    raise ValueError(f"Unsupported output format: {ext}")
+    
+  print(f"{output_file} created.")
 # end
 
 
@@ -204,7 +211,7 @@ def globalrange(data,kwargs):
 @click.option("--notitle", is_flag=True, help="Do not show title.")
 @click.option("-i", "--interval", default=100, help="Specify the animation interval.")
 @click.option("--save", is_flag=True, help="Save figure as PNG.")
-@click.option("--saveas", type=click.STRING, default=None, help="Name to save the plot as.")
+@click.option("--saveas", type=click.STRING, default=None, help="Name to save the plot as. Use .gif please.")
 @click.option("--fps", type=click.INT, help="Specify frames per second for saving.")
 @click.option("--dpi", type=click.INT, help="DPI (resolution) for output.")
 @click.option("-e", "--edgecolors", type=click.STRING, help="Set color for cell edges.")
@@ -230,6 +237,10 @@ def animate(ctx, **kwargs):
   """
   verb_print(ctx, "Starting animate")
   data = ctx.obj["data"]
+
+  if kwargs["saveas"] and not kwargs["saveas"].lower().endswith(".gif"):
+    raise click.ClickException("Currently only .gif output is supported for animations; please specify a .gif file with --saveas.")
+  # end
 
   if kwargs["xlim"]:
     kwargs["xmin"] = float(kwargs["xlim"].split(",")[0])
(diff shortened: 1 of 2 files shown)


src/postgkyl/commands/animate.py
```diff
@@ -159,7 +211,7 @@ def globalrange(data,kwargs):
 @click.option("--notitle", is_flag=True, help="Do not show title.")
 @click.option("-i", "--interval", default=100, help="Specify the animation interval.")
 @click.option("--save", is_flag=True, help="Save figure as PNG.")
-@click.option("--saveas", type=click.STRING, default=None, help="Name to save the plot as.")
+@click.option("--saveas", type=click.STRING, default=None, help="Name to save the plot as. Use .gif please.")
```
manauref: Rephrase "Use .gif please".
If the default is .gif, say "Default = movie.gif" or something like that.
If only gif is supported, say "Only .gif is allowed" or something like that.

However, I agree that .gif should be the default as that seems to be the most robust and more frequently used by people. But .mp4 used to work/works in main, so I think we should continue to allow that. If the user's environment doesn't work for .mp4 for some other reason, then the user can use .gif

src/postgkyl/commands/animate.py
```diff
@@ -169,19 +221,31 @@ def globalrange(data,kwargs):
 @click.option("--hashtag", is_flag=True, help="Turns on the pgkyl hashtag!")
 @click.option("--show/--no-show", default=True, help="Turn showing of the plot ON and OFF.")
 @click.option("--saveframes", type=click.STRING,
-    help="Save individual frames as PNGS instead of an animation")
+    help="Save individual frames as PNGs; also compiles a movie if --save or --saveas is given.")
```
manauref: Delete "; also compiles a movie if --save or --saveas is given"

options are generally combinable, so this is unnecessary.

src/postgkyl/commands/animate.py
```diff
@@ -169,19 +221,31 @@ def globalrange(data,kwargs):
...
+@click.option("--nproc", default=1, type=click.INT, show_default=True,
+    help="Number of parallel processes for frame generation. When >1 without --saveframes, "
```
manauref: I think "When >1 without --saveframes, frames are written to a temp directory, compiled into a movie, then removed"

should be removed. This is an implementation detail about the inner workings of postgkyl, not really relevant to the user. What I do think you might wanna do is give the command some kind of --saveframes_dir option, so the user can control where the temp directory is written (because it is in general bad practice to write files/folders without letting the user control where that happens).

src/postgkyl/commands/animate.py
```diff
@@ -169,19 +221,31 @@ def globalrange(data,kwargs):
...
+         "frames are written to a temp directory, compiled into a movie, then removed.")
 @click.option("--figsize", help="Comma-separated values for x and y size.")
 @click.option("-m", "--multiblock", is_flag=True, help="Plots blocks from each frame together")
 @click.pass_context
 def animate(ctx, **kwargs):
   """Animate the actively loaded dataset and show resulting plots in a loop.
 
   Typically, the datasets are loaded using wildcard/regex feature of the -f option to
-  the main pgkyl executable. To save the animation ffmpeg needs to be installed.
+  the main pgkyl executable. Saving via --saveframes or --nproc uses Pillow (GIF by
```
manauref: "uses Pillow (GIF by default); the interactive FuncAnimation path still requires ffmpeg."

1. should be deleted I think. "uses Pillow (GIF by default)" is an implementation detail, the user doesn't need to know about that.
2. This sentence is about --saveframes, so why is it then talking about GIF and FuncAnimation. 

The default and role of --saveas and --saveframes will already be clear in the description of these options. IT's best to keep descriptions very brief and without repetition.

src/postgkyl/commands/animate.py
```diff
@@ -223,6 +287,7 @@ def animate(ctx, **kwargs):
     figsize = (int(kwargs["figsize"].split(",")[0]), int(kwargs["figsize"].split(",")[1]))
   # end
 
+  duration = int(1000 / kwargs["fps"]) if kwargs["fps"] else kwargs["interval"]
```
manauref: what is this `1000` here? why 1e3 and not 1e2 or 1e4?

manauref (changes requested):

src/postgkyl/commands/animate.py
Antoinehoff: ok I removed it and will put back mp4 capabilities

src/postgkyl/commands/animate.py
Antoinehoff: if you provide --saveframes you can control where the directory is written. I don't know if we want to add another option for a directory that would be automatically created an removed, do we?

src/postgkyl/commands/animate.py
Antoinehoff: because PIL measures in miliseconds, I can put 1e3 if you want.

src/postgkyl/commands/animate.py
manauref: i think we do because the user may have restrictions on disk space (permissions or limited space), so you want to let the user pick where to put this temp stuff

src/postgkyl/commands/animate.py
manauref: ah ok, i didn't know that. Just add a comment like 
# Here 1000 because PIL measures in milisec.

Antoinehoff: done

manauref (approved):

PR #212, opened 2026-06-24
make the registry and quantities classes and not dict for more flexibility

Antoinehoff:
I also added a few utility functions which simplifies the gk_load_quantity function. 

This is related to #209

### DIFF
--- a/src/postgkyl/commands/gk_load_quantity.py
+++ b/src/postgkyl/commands/gk_load_quantity.py
@@ -1,6 +1,3 @@
-"""
-Loading pre-named gyrokinetic quantities.
-"""
 import click
 
 from postgkyl.utils.gk_quantities.registry import gk_quant_registry
--- a/src/postgkyl/commands/gk_load_quantity.py
+++ b/src/postgkyl/commands/gk_load_quantity.py
@@ -51,8 +51,8 @@ def _get_src_combo_and_frames(path, name, species, qattr, **kwargs):
   frames_avail: set[int] = set()
   combo_idx = 0
   # Check each combination of sources.
-  for cidx in range(len(qattr["source"])):
-    combo = qattr["source"][cidx]
+  for cidx in range(len(qattr.source)):
+    combo = qattr.source[cidx]
     # Check each source for this combo.
     for src in combo:
       if isinstance(src, str) and src in gk_geo_files:
@@ -201,7 +201,7 @@ def gk_load_quantity(ctx, **kwargs):
 
   if kwargs['qlist']:
     # Print accepted quantities and exit.
-    valid = sorted(gk_quant_registry.keys())
+    valid = gk_quant_registry.list()
     print(f"Available quantities: {', '.join(valid)}.")
     return
 
@@ -210,12 +210,12 @@ def gk_load_quantity(ctx, **kwargs):
   data = ctx.obj["data"]
   verb_print(ctx, f"Loading quantity '{quantity}' for {kwargs['name']}")
 
-  if quantity not in gk_quant_registry:
-    valid = sorted(gk_quant_registry.keys())
+  if not gk_quant_registry.has(quantity):
+    valid = gk_quant_registry.list()
     raise ValueError(f"Unknown quantity '{quantity}'. "
                      f"Available quantities: {', '.join(valid)}.")
 
-  quant_attr = gk_quant_registry[quantity]
+  quant_attr = gk_quant_registry.get(quantity)
 
   # Parse --extra into a dict, auto-converting numeric values.
   user_extra = {}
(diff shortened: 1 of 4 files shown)


src/postgkyl/commands/gk_load_quantity.py
```diff
@@ -233,47 +77,31 @@ def gk_load_quantity(ctx, **kwargs):
...
-      label_tmpl = quant_attr["label"]
-      default_label = ''
-      if "%s" in label_tmpl:
-        if quantity == "ExB_vel":
-          default_label = label_tmpl % str(user_extra["dir"])
-        else:
-          default_label = label_tmpl % species[0]
-      else:
-        default_label = label_tmpl
+      default_label = gkquant.get_label(species=species[0], direction=user_extra.get("dir", ''))
```
manauref: I think there's an error in the label handling. Try loading geo_int_b_i and you'll see.
By the way: I did try all the quantities that i had implemented at some point (though later messed it up clearly), so would be good if you do that here too.

Somewhat related: one should be able to load b_i. I think -q b_i should load all components (because some calculations like ExB_vel need several components), and the user can later `select` a specific component.

Another option is to have a better fetch function for b_i, which allows the use of `--extra dir=` to select a specific component. What do you think?

src/postgkyl/utils/gk_quantities/gkquantity.py
```diff
@@ -0,0 +1,221 @@
...
+  def get_label(self, species : str | None = None, direction : str | None = None) -> str:
+    """Get the label for the quantity, replacing %s with species name or direction."""
+    if self.is_vector and direction is not None:
+      return self.label % str(direction)
+    elif self.is_species_dep and species is not None:
+      return self.label % species[0]
+    else:
+      return self.label
+
+  def src_stem(self, path : str, name : str, species : str, src : str) -> str:
```
manauref: should be `get_src_stem` i think
it's best to have actions for the function names of a class I think

src/postgkyl/utils/gk_quantities/gkquantity.py
```diff
@@ -0,0 +1,221 @@
...
+    """Stem of the file name for a string source, including the trailing
+    separator before the frame number (geo files have no frame, so no separator)."""
+    if self.is_geo:
+      return os.path.join(path, f"{name}-{src}")
+    elif self.is_species_dep:
+      return os.path.join(path, f"{name}-{species}_{src}_")
+    else:
+      return os.path.join(path, f"{name}-{src}_")
+
+  def src_file_name(self, path : str, name : str, species : str, src : str,
```
manauref: -> get src_file_name

src/postgkyl/utils/gk_quantities/gkquantity.py
```diff
@@ -0,0 +1,221 @@
...
+                    frame : int | None) -> str:
+    """Full file name for a string source at the given frame."""
+    if self.is_geo:
+      return f"{self.src_stem(path, name, species, src)}.gkyl"
+    else:
+      return f"{self.src_stem(path, name, species, src)}{frame}.gkyl"
+
+  def avail_frames_src(self, path : str, name : str, species : str, src : str,
```
manauref: -> get_available_frames_src

src/postgkyl/utils/gk_quantities/gkquantity.py
```diff
@@ -0,0 +1,221 @@
...
+    else:
+      candidates = glob.glob(f"{glob.escape(stem)}*.gkyl")
+
+    for f in candidates:
+      suffix = f[len(stem):-5]
+      if suffix.isdigit():
+        frames_avail.add(int(suffix))
+    return frames_avail
+
+  def avail_combo_frames(self, path : str, name : str, species : str,
```
manauref: -> find_avail_combo_frames

manauref (approved):
approving, but made a couple of comments to be addressed before merging.

src/postgkyl/utils/gk_quantities/gkquantity.py
Antoinehoff: Let me just add "_" because this is not meant to be an interface

src/postgkyl/utils/gk_quantities/gkquantity.py
Antoinehoff: Let me just add "_" because this is not meant to be an interface

src/postgkyl/utils/gk_quantities/gkquantity.py
Antoinehoff: Let me just add "_" because this is not meant to be an interface

src/postgkyl/utils/gk_quantities/gkquantity.py
Antoinehoff: Let me just add "_" because this is not meant to be an interface

src/postgkyl/commands/gk_load_quantity.py
Antoinehoff: ok I fixed it and now this command
```
pgkyl gk-load-quantity --quantity geo_int_b_i --name rt_gk_helical_zpar_adapt_src_1x2v_p1 --frame 0 --path /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_helical_1x2v.o//wk_sgbc_maxgauss_ic/ interpolate plot
```
produces three subplots, one for each component. I don't understand how to use the label we created above in this plot though.

PR #213, opened 2026-06-24
new command DG local poly

Antoinehoff:
This command is similar to the `interp` one, i.e. it evaluates the active DG array onto points. The specificity here is that it will evaluate the field on `npointspercell` points within each cell, that are contained between `-1+eps` and `1-eps` in the logical coordinates.
`eps` and `npointspercell` are options to the command.
Each sequence of values are separated by one NaN value at the cell interface to cut the curves when plot.

### p=1 1D

```
pgkyl /rt_gk_helical_zpar_adapt_src_1x2v_p1-elc_M0_1.gkyl dg-local-poly plot --figure 0 --xlabel '$z$' --subplot-titles rt_gk_helical_zpar_adapt_src_1x2v_p1 --subplot-ylabels elc_M0 --no-showgrid
```

image

or with time evolution:

```
pgkyl /rt_gk_helical_zpar_adapt_src_1x2v_p1-elc_M0_*.gkyl dg-local-poly collect plot --figure 0 --surface --xlabel '$z$' --subplot-titles rt_gk_helical_zpar_adapt_src_1x2v_p1 --subplot-ylabels elc_M0 --no-showgrid
```
image

### p=2 1D

```
pgkyl rt_vlasov_weibel_1x2v_p2-elc_M0_1.gkyl dg-local-poly --pointspercell 16 plot --figure 0 --xlabel '$z$' --subplot-titles rt_vlasov_weibel_1x2v_p2 --subplot-ylabels elc_M0 --no-showgrid
```
image

### p=2 2D

```
pgkyl rt_vlasov_weibel_2x2v_p2-elc_M0_1.gkyl dg-local-poly --pointspercell 3 plot --figure 0 --surface --clabel elc_M0 --subplot-titles rt_vlasov_weibel_2x2v_p2 --subplot-xlabels '$z$' --no-showgrid --xkcd --colormap jet
```

image


src/postgkyl/commands/dg_local_poly.py
```diff
@@ -0,0 +1,101 @@
...
+from postgkyl.data.dg import _getnum_nodes
+from postgkyl.modalDG.kernels import expand_1d, expand_2d, expand_3d
+
+
+@click.command()
+@click.option("--use", "-u", help="Specify a 'tag' to apply to (default all tags).")
+@click.option("--eps", type=click.FLOAT, default=1e-2,
+    help="Fraction of the half-cell width by which the evaluation point is "
+         "moved inside from each cell interface (must be in [0, 1]).")
+@click.option("--pointspercell", type=click.INT, default=2,
```
manauref: just call this --npoints for simplicity, The description makes it clear. Also give it the -n option.

src/postgkyl/commands/dg_local_poly.py
```diff
@@ -0,0 +1,101 @@
...
+import numpy as np
+
+from postgkyl.utils import verb_print
+from postgkyl.data.dg import _getnum_nodes
+from postgkyl.modalDG.kernels import expand_1d, expand_2d, expand_3d
+
+
+@click.command()
+@click.option("--use", "-u", help="Specify a 'tag' to apply to (default all tags).")
+@click.option("--eps", type=click.FLOAT, default=1e-2,
```
manauref: give it the -e option

also, what is the reason for this epsilon? why can't you go right up to -1 and 1? especially if you are putting a NaN between cells

I tried this command and because of this eps, I think, there's a noticeable gap between cells (when you zoom in). I think we don't want this. We want to plot the absolute true DG representation. So I think it's better if the default eps=0 (but you can leave it as an option if people want that gap).

src/postgkyl/commands/dg_local_poly.py
```diff
@@ -0,0 +1,101 @@
...
+    help="Fraction of the half-cell width by which the evaluation point is "
+         "moved inside from each cell interface (must be in [0, 1]).")
+@click.option("--pointspercell", type=click.INT, default=2,
+    help="Number of evaluation points per cell.")
+@click.pass_context
+def dg_local_poly(ctx, **kwargs):
+  """
+  Generate a discontinuous DG polynomial cellwise representation of the data.
+  The modal DG decomposition is evaluated at two points per cell, each located
+  just inside a cell interface (slightly interior, controlled by --eps). A NaN
```
manauref: this comment "two points per cell" seems specific to 1D p1. I'd remove it or generalize it.

src/postgkyl/commands/dg_local_poly.py
```diff
@@ -0,0 +1,101 @@
...
+    if poly_order is None:
+      ctx.fail(click.style(
+          "ERROR in dg-local-poly: no 'poly_order' was specified and dataset "
+          f"{dat.get_label():s} does not have the required information.",
+          fg="red"))
+
+    num_dims = dat.get_num_dims()
+    if num_dims > 3:
+      ctx.fail(click.style(
+          "ERROR in dg-local-poly: only data with up to 3 dimensions is supported.",
```
manauref: 3D datasets are supported? how? what does it do for a 3D dataset?

src/postgkyl/modalDG/kernels/expand1d.py
```diff
@@ -7,7 +7,7 @@ def _expand_1d1p(f, x):
 
 def _expand_1d2p(f, x):
   return (
-      2.371708245126284 * f[..., 2] * (x ^ 2 - 0.3333333333333333)
+      2.371708245126284 * f[..., 2] * (x ** 2 - 0.3333333333333333)
```
manauref, line 10: why do you have to change this? it's been working this whole time

.gitignore
```diff
@@ -4,3 +4,5 @@ dist/
 postgkyl.egg-info/
 src/postgkyl/version.py
 .DS_Store
+.venv
+.new_venv
```
manauref: what are these?

manauref (approved):
very nice and valuable addition. Just made a couple of comments that would be good to address before merging

src/postgkyl/commands/dg_local_poly.py
Antoinehoff: Yeah I don't think this is necessary indeed, I think it was related to an old method I was doing but with the NaN we don't need it.

src/postgkyl/commands/dg_local_poly.py
Antoinehoff: done

src/postgkyl/commands/dg_local_poly.py
Antoinehoff: yep and I removed also the eps

src/postgkyl/commands/dg_local_poly.py
Antoinehoff: I mean this function only evaluates on points and add "layers" of NaNs. There is no dimensionality limitation on this method, though it must be extremely costly to apply it to a 5D array. I think I should limit it to 2D.

manauref: i see your point. I guess it doesn't matter, and the user may be allowed to live dangerously if that's what they choose

Antoinehoff: ok let's give up to 6D then lol

src/postgkyl/modalDG/kernels/expand1d.py
Antoinehoff: I don't think it was used. Also if you compare with the expand_2d, it does not make sense. The expand_2d does not use ^ but ** everywhere. Don't ask me why but this is safer.

.gitignore
Antoinehoff: arf sorry I commited something I shouldn't

Issue #201, opened 2026-06-15
[DR] gk-rz command for plotting on the R-Z plane


manauref:
An operation we very frequently need to do with Gk data is plot quantities on the R-Z plane. We propose to simplify this by introducing a `gk-rz` command that performs this transformation. It'll simply work as
```
pgkyl gk_ltx_iwl_2x2v_p1-elc_M2par_10.gkyl gk-rz --mapc2p gk_ltx_iwl_2x2v_p1-mapc2p.gkyl pl -a
```
to produce

<img width="752" height="656" alt="Image" src="https://github.com/user-attachments/assets/77d8decc-650b-4abc-b6c6-16945247ca30" />

(of course this is rather trivial in 2D and can already be done passing --c2p to the data loading, but it's more complex in 3D and multiblock, so gk-rz will black-box all the complexity away from users).

Note that the data loading was carried out prior to gk-rz. Thus all gk-rz will do is:
- Obtain the R-Z-phi coordinates from `-mapc2p.gkyl`
- Interpolate along the z direction if needed.

It'll optionally take a higher resolution `-mapc2p.gkyl` file so plots look smoother.

This is being prototyped in the `gk_rz_plot` branch.

PR #214, opened 2026-06-25
RZ plot for GK tokamak data

Antoinehoff:
This PR is related to DR #201

It implements the projection of tokamak data onto $R$-$Z$ and $\theta$-$\varphi$ planes.

### gk-rz
Here is an example of the new `gk-rz` command. If we do not pass a path to `mapc2p` or `nodes` files, it will automatically look for the `nodes` file using the same prefix as the current data in the stack to perform the `xyz` to `RZ` mapping. One can choose the toroidal angle using `--phi-tor` option.
```shell
pgkyl gk-load-quantity --quantity M0 --name rt_gk_tcv_nt_iwl_3x2v_p1  --species elc --frame 5 gk-rz --phi-tor 1.5 pl -a &
```

Screenshot 2026-07-08 at 10 02 25 AM

### gk-fluxsurf
Given a radial index `-x`, `gk-fluxsurf` uses `gk-rz` to get perimeters of data at constant $r$ and $\varphi$, i.e. the field is expressed as $f=f(\theta)$ only. Then, it stitches many of these perimeters together for many $\varphi$ values to obtain a $\theta,\varphi$ surface plot. On that plane, the turbulent structures are aligned with the magnetic field lines and we can appreciate the local q-profile and other magnetic geometry features.

```
pgkyl gk-load-quantity --quantity M0 --name rt_gk_tcv_nt_iwl_3x2v_p1  --species elc --frame 5 gk-fluxsurf -x 10 pl
```

Screenshot 2026-07-08 at 10 05 26 AM


manauref:
Here are a few things I think need to be addressed:
1. Right now there could be a datasets with several components loaded, but gk-rz just plots the first of them. For example
```
pgkyl rt_gk_tcv_iwl_adapt_source_3x2v_p1-elc_BiMaxwellianMoments_0.gkyl gk-rz pl -a
```
only plots the density. I'd think it should plot all 4 components in subplots like plot does.

2. I tried it with a 2x MAPC2P input file, for example
```
pgkyl rt_gk_tcv_iwl_adapt_source_2x2v_p1-elc_BiMaxwellianMoments_0.gkyl gk-rz pl -a
```
and a significant gap remains at the IMP:
Screenshot 2026-07-11 at 9 03 58 PM
maybe some periodicity needs to be applied?

3. The issue in 2 above is particularly bad for a 2D TOKAMAK geo LTX input file. The following
```
pgkyl gk_ltx_iwl_2x2v_p1-elc_BiMaxwellianMoments_2.gkyl gk-rz pl -a
```
gives
Screenshot 2026-07-11 at 9 28 36 PM

I still doubt whether using geo_int_nodes.gkyl is the correct thing to do. Note that if one instead uses the already existent --c2p with -geo_corn_mapc2p_deflated.gkyl, the plot has a much smaller gap at the IMP. That is:
```
pgkyl --c2p gk_ltx_iwl_2x2v_p1-geo_corn_mapc2p_deflated.gkyl gk_ltx_iwl_2x2v_p1-elc_BiMaxwellianMoments_2.gkyl sel -c :4 interp pl -a
```
gives
Screenshot 2026-07-11 at 9 40 29 PM

I know interpolating geometry to nodes where they are not meant to be evaluated has given inaccurate plots in the past, but could we use both -geo_corn_mapc2p.gkyl and -geo_int_mapc2p.gkyl to provide a better interpolation?

4. A question: can we use --mapc2p or --nodes to pass a higher resolution file to make the plot look smoother?

manauref (approved):
please add a test for gk_fluxsurf if you can

PR #216, opened 2026-06-26
An alternative implementation of selecting negative indicies

Maxwell-Rosen:
This implementation wraps the input index around into the appropriate end of the array. This can accommodate commands like `pgkyl file.gkyl sel --z0 -1`.

### DIFF
--- a/src/postgkyl/data/idx_parser.py
+++ b/src/postgkyl/data/idx_parser.py
@@ -24,7 +24,7 @@ def _find_cell_index(array, value):
 
 def _string_to_index(value: str, array: np.ndarray, nodal: bool = False) -> int:
   if isinstance(value, str):
-    if value.isdigit():
+    if value.lstrip("-").isdigit():
       return int(value)
     else:
       if nodal:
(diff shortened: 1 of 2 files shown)


src/postgkyl/data/idx_parser.py
```diff
@@ -24,7 +24,7 @@ def _find_cell_index(array, value):
 
 def _string_to_index(value: str, array: np.ndarray, nodal: bool = False) -> int:
   if isinstance(value, str):
-    if value.isdigit():
+    if value.lstrip("-").isdigit():
```
Antoinehoff, line 27: This is better than my `try` command I agree

src/postgkyl/data/select.py
```diff
@@ -55,6 +55,10 @@ def select(data: GData, comp: int | str | None = None,
       is_matching = values.shape[d] == len_grid
       idx = idx_parser.idx_parser(z, grid[d], is_matching)
       if isinstance(idx, int):
+        axis_cells = values.shape[d]
+        if idx < 0: # Wrap negative index around
+          idx = axis_cells + idx
```
Antoinehoff, line 60: I don't think select should manipulate the index, it's better to have idx_parser do everything and just get the correct index.

manauref:
@Antoinehoff learned from you and modified his PR. This will likely change in pgkyl v2 anyway

Issue #208, opened 2026-06-23
[DR] dg-evproj: evaluate a DG field and project onto lower dim basis


manauref:
Summary
----------

We propose to add a command, named dg-evproj, that:
1. Evaluates the (already) loaded DG field at a specific coordinate if it is given as float, or at a cell center if given as an integer.
2. Projects onto the lower dimensional basis.

The second step is so that subsequent commands see a DG field whose dimensionality is consistent with its normalization (and thus something like `interp` works correctly).

For example::
```
pgkyl rt_gk_sheath_2x2v_p1-field_1.gkyl dg-evproj --z1 -3.141593 
```
will evaluate the field at z1=-pi, and produce a 1D DG field (with 2 coefficients per cell) with the normalization of a 1D basis function. One can subequently do something like
```
pgkyl rt_gk_sheath_2x2v_p1-field_1.gkyl dg-evproj --z1 -3.141593 interp pl
```
and the 1D plot will work as expected.

Note that there's a task here of given a coordinate, find the cell that owns that coordinate. We'll do that by calling `select` under the hood at first. At cell boundaries `select` defaults to using the expansion in the lower cell; we'll let the user force selection of the expansion at the upper cell with the `--use_up` option:
```
pgkyl rt_gk_sheath_2x2v_p1-field_1.gkyl dg-evproj --z1 0.0 --use-up interp pl
```
plots the x variation of the field at z=0 using the DG expansion on the z>0 side of z=0.

Projection via Gkeyll
---------------------

The projection step is better handled inside Gkeyll, because gkeyll knows about basis types and normalizations and we'd rather not hand-transfer that info into postgkyl.

The updater that most closely resembles this operation, to my knowledge, is `gyrokinetic/zero/gkyl_translate_dim_new.h`. The only hiccup is that this is only meant to evaluate at logical -1, 0 or +1, but I think we can easily extend this updater to take an arbitrary logical coordinate (and move it into core/ so that postgkyl only depends on the core lib).

Then we'll call `gkyl_translate_dim_new` from postgkyl like we already call weak division, multiplication and differentiation in the `dg_binop_sig_change` and `dg_binop_sig_change-differentiate` branches.

PR #217, opened 2026-06-29
Command evaluating DG dataset at coord & project onto lower-dim basis

manauref:
## Summary

Per DR in issue #208:

We implement a new command which exactly evaluates a loaded DG field at a given coordinate, and projects it onto a lower-dimensional basis, returning the result as a GData pushed to the stack so that subsequent pgkyl commands can operate on it.

This depends on
- [PR 1071 of gkeyll](https://github.com/ammarhakim/gkeyll/pull/1071).
- [PR 102 of gkylcas](https://github.com/ammarhakim/gkylcas/pull/102).

I validated the result of this capability against our previous workflow (`pgkyl <file> interp select <ops> plot`) by using the new command to evaluate a the exact same coordinates that `interp` would.

Something to comment on:
For loading, evaluating and plotting distributions (see example below) we modified `gk-distf` to allow for `-i 0`, in which case:
- No interpolation takes place.
- Divides by jacobvel.
- Does weak multiplication by jacobtot_inv
- Pushes the resulting DG field to stack.
This way, when followed by `dg-evproj`, we can evaluate the distribution exactly at a coordinate. Interpolation is later handled with an interpolating command like `interp` or `dg-local`. ** I wonder ** whether the default should be `-i 0`, and the user has to pass `-i` or `-i N` if they really want to interpolate.

## Examples

Conf-space example: evaluating $u_{\parallel}$ at the sheath entrance
```
pgkyl rt_gk_tcv_iwl_adapt_source_2x2v_p1-elc_BiMaxwellianMoments_1.gkyl \
  dg-evproj -c1 --z1 -3.141593 \
  pl -x '$x$ (m)' -y '$u_{\parallel e}(z=-\pi)$' --title 'rt_gk_tcv_iwl_adapt_source_2x2v_p1-elc_BiMaxwellianMoments_1.gkyl'
```
Screenshot 2026-06-29 at 11 44 27 AM

Vel-space example: evaluating the distribution at z=0 (combined with the new `dg-local-poly` command)
```
pgkyl gk-distf -n rt_gk_sheath_nonuniformv_1x2v_p1 -s elc -f 1 -i 0 --c2p-vel \
  dg-ev --z0 -0.125 --z1 0. dg-local \
  pl -x '$\mu$' -y '$f_e$' --title 'rt_gk_sheath_nonuniformv_1x2v_p1-elc_1.gkyl'
```
Screenshot 2026-06-29 at 11 42 41 AM


src/postgkyl/commands/gk_distf.py
```diff
@@ -98,67 +76,74 @@ def load_gk_distf(
...
-  out = GData(tag=tag, ctx=jf_data.ctx)
+  out = GData(tag=tag, ctx=Jf_data.ctx)
   out.push(out_grid, f_values)
   return out
 # end
 
-# Generated by LLMs, commented and verified by MR 3/16/26
 @click.command()
 @click.option("--name", "-n", required=True, type=click.STRING,
   help="Simulation name prefix (e.g. gk_lorentzian_mirror).")
```
Antoinehoff, line 141: Is this comment correct? It seems that we need to be in the directory for it to work then. I think it would be good if it works with a path too or accept the path somewhere else.

manauref: I think the comment is correct, although "prefix" could mean many rhings. Maybe it should just say "name"?

gk-distf doesn't have a -p --path option, but we could add it

Antoinehoff: I tested actually and we can add the path to prefix and it works. maybe the comment can say `e.g. path/gk_lorentzian_mirror` idk

PR #219, opened 2026-06-30
Add an interface to use gk-distf through the gk-load-quantity.

Antoinehoff:
These changes allow us to write :
```
pgkyl gk-load-quantity -q distf -n rt_gk_tcv_iwl_adapt_source_1x2v_p1 -s elc -f 0
```
i.e. to use gk-loadq for distribution function as well. 
This is done by adding a new `loader_func` attribute to `GkQuantity` which allows to have an alternative way to load Gdata.
We can then add to the registry:
```Python
# Distribution function loaded through load_gk_distf.
_distf : GkQuantity = GkQuantity(
  name = "distf",
  source = [[""]],
  fetch_func = [None],
  loader_func = ff.load_distf,
  label = r"$f_{%s}$",
  is_time_dep = True,
  is_species_dep = True,
)
gk_quant_registry.register(_distf)
```
Where `ff.load_distf` is a small wrapper defined with the other fetch functions in `fetch_funcs.py`. This wrapper allows us to define some flags by default. For example, here we decide to always load the c2p_vel. We can modify that with the `--extra` command.

### DIFF
--- a/src/postgkyl/utils/gk_quantities/fetch_funcs.py
+++ b/src/postgkyl/utils/gk_quantities/fetch_funcs.py
@@ -528,3 +528,35 @@ def fetch_diamag_vel(gdatas, **kwargs):
   out.set_values(out.get_values()/charge)
 
   return out
+
+# -------------------------------------------------------------------------
+# --- Custom loaders (loader_func signature: path, name, species, frame) ---
+# -------------------------------------------------------------------------
+
+def load_distf(path: str, name: str, species: str, frame: int, **extra) -> GData:
+  """
+  Loader for the registry 'distf' quantity. Wraps load_gk_distf with defaults
+  tailored to registry use: never interpolate (interp=0) and convert velocity
+  coordinates (c2p_vel) on by default.
+
+  Defaults can be overridden via --extra, e.g.:
+    -e suffix=source      use <name>-<species>_source_<frame>.gkyl as input
+    -e c2p_vel=0          disable velocity-space mapping
+    -e mc2nu=1            apply non-uniform -> field-aligned position mapping
+    -e mapc2p=1           apply position-space -> Cartesian/cylindrical mapping
+    -e block=2            load only the 2nd block of a multi-block file
+  """
+  from postgkyl.commands.gk_distf import load_gk_distf
+  from postgkyl.utils.gk_utils import dict_get_bool
+
+  prefix = path.rstrip("/") + "/" + name
+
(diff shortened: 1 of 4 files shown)


src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -527,4 +527,36 @@ def fetch_diamag_vel(gdatas, **kwargs):
   charge = _get_ctx_val(pressperp, "charge", **kwargs)
   out.set_values(out.get_values()/charge)
 
-  return out
\ No newline at end of file
+  return out
+
+# -------------------------------------------------------------------------
+# --- Custom loaders (loader_func signature: path, name, species, frame) ---
+# -------------------------------------------------------------------------
```
manauref: remove, Keep comments simple with style already in codebase

maybe it's time for a SKILL.md

src/postgkyl/utils/gk_quantities/gkquantity.py
```diff
@@ -15,10 +15,12 @@ class GkQuantity:
     is_time_dep: If the quantity is time-dependent (i.e. written in frames).
     is_species_dep: If the quantity is species-dependent.
     is_vector: If the quantity is a vector (i.e. has multiple components).
+    loader_func: Optional custom loader which bypasses the fetch machinery.
   """
   name = None
   source = None
   fetch_func = None
+  loader_func = None
```
manauref: use fetch function instead, no need for loader func

src/postgkyl/utils/gk_quantities/registry.py
```diff
@@ -283,4 +283,20 @@
   is_species_dep = True,
   is_vector = True
 )
-gk_quant_registry.register(_diamag_vel)
\ No newline at end of file
+gk_quant_registry.register(_diamag_vel)
+
+# ------------------------------
+# --- Phase space quantities ---
+# ------------------------------
```
manauref, line 290: remove

keep comment simple with existing style

manauref (changes requested):

src/postgkyl/utils/gk_quantities/registry.py
Antoinehoff: this fits existing style, I like to have this so that we remember that there is an organization on how we list the registry. Otherwise it will become messy.

manauref (approved):

PR #222, opened 2026-07-07
DG average operator

Antoinehoff:
We add the `dg-avg` command to get the average (using the jacobian by default) over that direction. For example
```
pgkyl rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_5.gkyl dg-avg --z1 --z2
```
is the flux surface average. It is still in DG representation but now it is 1D.
The implementation is similar to the `dg-ev` command (see #217), we just use the `array_average` updater from Gkeyll through python.

We need to merge https://github.com/ammarhakim/gkeyll/pull/1081 for this new command to work.

### Examples
We compare 
```
pgkyl rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_5.gkyl dg-ev --z1 0.0 interp pl
```
with 
```
pgkyl rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_5.gkyl dg-avg --z1 interp pl
```

below is both outputs, `z1=0.0` on the left, `z1=avg` on the right.

Screenshot 2026-07-07 at 12 16 38 PM

We can also combine like
```
pgkyl rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_5.gkyl dg-avg --z1 dg-ev --z1 0.0 interp pl
```
Note here that the first `--z1` option is not the same direction as the second `--z1` option since the data is reduced by the `dg-ev` command.

Screenshot 2026-07-07 at 12 21 17 PM

### Caveat
The `dg-avg` looks automatically for a jacobian file to use according to the prefix of the file on the stack. This works well until we use something like `ev` which will not pass the file name information further, preventing the search of the jacobian file.
For example:
```
pgkyl -v rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_5.gkyl rt_gk_tcv_nt_iwl_3x2v_p1-ion_M0_4.gkyl ev 'f[1] f[0] -' dg-avg --z2 interp pl
```
will choose the weightless average
```
[0.003136] Starting dg-avg
[0.088369] No jacobgeo weight found (None); using unweighted average.
[0.088724] Finishing dg-avg
```


manauref:
@Antoinehoff and I discussed. It seems better to put this into a separate command for now, so that dg-ev remains a wrapper for gkeyll's dg_eval_at_coord_proj operator, and the new dg-average command will likewise be a wrapper for gkeyll's DG average.

In the future we may use other commands that can wrap several commands (e.g. maybe the existing `integrate` could call dg_average or do what it currently does on interpolated data)

Antoinehoff:
> @Antoinehoff and I discussed. It seems better to put this into a separate command for now, so that dg-ev remains a wrapper for gkeyll's dg_eval_at_coord_proj operator, and the new dg-average command will likewise be a wrapper for gkeyll's DG average.
> 
> In the future we may use other commands that can wrap several commands (e.g. maybe the existing `integrate` could call dg_average or do what it currently does on interpolated data)

Done

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -21,6 +21,50 @@
 # gkyl_elem_type enum ordinal for double (INT=0, FLOAT=1, DOUBLE=2)
 _GKYL_DOUBLE = ctypes.c_int(2)
 
+
+class _GkylBasis(ctypes.Structure):
```
manauref: we are using opaque wraps of gkeyll objects. Please delete this and use existing methods for creating a basis.

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -21,6 +21,50 @@
...
+    ("nodal_to_modal_quad_surf", ctypes.c_void_p * 3),
+    ("node_quad_surf_list",      ctypes.c_void_p * 3),
+    ("quad_nodal_to_modal",      ctypes.c_void_p),
+    ("modal_to_quad_nodal",      ctypes.c_void_p),
+  ]
+
+
+class _GkylArrayAverageInp(ctypes.Structure):
+  """ctypes mirror of `struct gkyl_array_average_inp` (core/zero/gkyl_array_average.h)."""
+  _fields_ = [
```
manauref: we are using opaque wraps, so we do not tell python the contents of any gkeyll struct. Only the types of the arguments to functions.

so please delete this and follow the pattern for wrapping other updaters.

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -70,6 +114,10 @@ def _setup_signatures(self) -> None:
     lib.gkyl_cart_modal_gkhybrid_new.argtypes = [c_i, c_i]
     lib.gkyl_cart_modal_gkhybrid_new.restype  = c_vp
 
+    # gkyl_cart_modal_serendip(*basis, ndim, poly_order) -> void (fills in place)
+    lib.gkyl_cart_modal_serendip.argtypes = [ctypes.POINTER(_GkylBasis), c_i, c_i]
+    lib.gkyl_cart_modal_serendip.restype  = None
```
manauref: delete

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -134,6 +182,18 @@ def _setup_signatures(self) -> None:
     lib.gkyl_dg_eval_at_coord_proj_release.argtypes = [c_vp]
     lib.gkyl_dg_eval_at_coord_proj_release.restype  = None
 
+    # gkyl_array_average_new(inp*) -> gkyl_array_average*
+    lib.gkyl_array_average_new.argtypes = [ctypes.POINTER(_GkylArrayAverageInp)]
```
manauref: just list the types like the other updaters

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -385,6 +445,153 @@ def eval_at_coord_proj(self, eval_dirs: list, eval_coords: list, gdata,
...
+
+    Inputs:
+      avg_dirs:  List of 0-based direction indices to average over.
+      gdata:     Donor DG dataset (serendipity, poly_order in ctx).
+      weight:    Optional weight GData, same dimensionality/basis as gdata.
+                 If None, an unweighted average (integral / volume) is done.
+      comp_grid: Passed to the output GData constructor.
+
+    Returns:
+      GData with the averaged field over the surviving grid dimensions.
```
manauref: way too verbose. Just simplify it to a few lines (see other functions below). If more detail is needed, the user can look at gkeyll.

The reason is that this is that in general comments are not well maintained, and long comments even less. If we change something in gkeyll, a long comment like this is less likely to be updated. And it's preferable to have a better comment in gkeyll than postgkyl.

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -385,6 +445,153 @@ def eval_at_coord_proj(self, eval_dirs: list, eval_coords: list, gdata,
...
+    c_rng_up_tar = (ctypes.c_int * ndim_red)(*cells_tar)
+    rng_tar_ptr  = self._lib.gkyl_range_new(ctypes.c_int(ndim_red), c_rng_lo_tar, c_rng_up_tar)
+
+    # Full and reduced serendipity bases (filled in place, no release needed).
+    basis     = _GkylBasis()
+    basis_avg = _GkylBasis()
+    self._lib.gkyl_cart_modal_serendip(ctypes.byref(basis),
+      ctypes.c_int(ndim), ctypes.c_int(poly_order))
+    self._lib.gkyl_cart_modal_serendip(ctypes.byref(basis_avg),
+      ctypes.c_int(ndim_red), ctypes.c_int(poly_order))
```
manauref: use methods that already existed for creating basis

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -385,6 +445,153 @@ def eval_at_coord_proj(self, eval_dirs: list, eval_coords: list, gdata,
...
+    tar_shape      = (*cells_tar, ncomp_tar)
+    tar_buf        = np.zeros(tar_shape, dtype=np.float64)
+    arr_tar        = self._lib.gkyl_array_new_from_buff(_GKYL_DOUBLE, ctypes.c_size_t(ncomp_tar),
+      ctypes.c_size_t(size_tar), tar_buf.ctypes.data_as(ctypes.c_void_p), )
+
+    # avg_dim flags (1 = averaged) over the full dimensionality.
+    avg_flags   = [1 if d in avg_dirs else 0 for d in range(ndim)]
+    c_avg_dim   = (ctypes.c_int * ndim)(*avg_flags)
+
+    inp = _GkylArrayAverageInp()
```
manauref: call the _new function here, as we would in gkeyll and as we do with other wrapped updaters

src/postgkyl/commands/dg_avg.py
```diff
@@ -0,0 +1,101 @@
...
+      return None
+    if weight_opt:  # explicit override: must exist
+      path = weight_opt
+      if not os.path.isfile(path):
+        ctx.fail(f"weight file '{path}' not found.")
+    else:  # auto-detect the geometric Jacobian
+      path = _jacobgeo_path(dat.get_file_name())
+      if path is None or not os.path.isfile(path):
+        verb_print(ctx, f"No jacobgeo weight found ({path}); using unweighted average.")
+        return None
```
manauref, line 76: I'm not sure this should be here. On the one hand this is meant for a flux-surface average, which there may be a separate command for. On the other hand, averaging in curvilinear coordinates always necessitates the Jacobian.

Perhaps more importantly: getting the jacobian from the context (get_file_name) will not work for derived datasets whose contexts may not have this quantity. Maybe it's ok, and for those cases one must specify it manually, but I'm not sure.

manauref (changes requested):

src/postgkyl/commands/dg_avg.py
Antoinehoff: I think the derived dataset now have at least one filename attached to it, if you are talking about `gk-load`. I tried and it works. I just don't want to have to pass the jacobian path manually, it's annoying especially because in 99% of usage case, we will use that one.

manauref: well derived can be from any number of operations, not just gk-load.
Try using `ev` to add two datasets and then average the resulting dataset

manauref (approved):

src/postgkyl/commands/dg_avg.py
Antoinehoff: Ok I add a caveat to the PR description

PR #224, opened 2026-07-11
Marimo notebook for visualizing DG twistshift

manauref:
This depends on [PR 1084 of gkeyll](https://github.com/ammarhakim/gkeyll/pull/1084).

We add a marimo notebook to visualize the DG treatment of twistshift for a specific input file. That is, this notebook:
- Compiles and runs (the initialization of) a specific GK input file.
- Plots the region covered by the lower-y cells shifted according to the shift set by the input file.
- Plots a specific quantity, and its twistshifted version.


Antoinehoff:
It can also apply TS on a .gkyl file we provide right?

manauref:
no. You specify an input file, and a quantity written out by that input file. Not a .gkyl file.

Antoinehoff:
ok, I guess I could make a restart and then be able to analyze the frame I want like this.

Antoinehoff (approved):

PR #225, opened 2026-07-13
Fix transpose plot option

Antoinehoff:
The transpose option was doing nothing in plot command. It nows swtiches the horizontal and vertical axis.

Examples
```
pgkyl gk-load-quantity --quantity distf --name rt_gk_tcv_iwl_csbc_1x2v_p1 --species elc --frame 10 --path /Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk/ interp select --z0 -3.1415 pl --transpose
```
in *main* returns:

Screenshot 2026-07-13 at 3 20 39 PM

With this PR it returns

Screenshot 2026-07-13 at 3 20 11 PM

Same for 1D plots:

```
pgkyl tests/test_data/rt_gk_tcv_iwl_1x2v_p1-elc_250.gkyl interp select --z0=2 --z2=0 pl --transpose 
```
*main*:
Screenshot 2026-07-13 at 3 22 05 PM

this branch:
Screenshot 2026-07-13 at 3 22 30 PM

### DIFF
--- a/src/postgkyl/output/plot.py
+++ b/src/postgkyl/output/plot.py
@@ -135,7 +135,7 @@ def plot(data: GData | Tuple[list, np.ndarray], args: list = (),
     mpl.rcParams["lines.linestyle"] = linestyle
   # end
 
-  # ---- Data Loading ----
+  # Data Loading
   # Get the handles on the grid and values
   grid_in, values = input_parser(data)
   grid = grid_in.copy()
@@ -197,7 +197,6 @@ def plot(data: GData | Tuple[list, np.ndarray], args: list = (),
     # end
   # end
 
-  # ---- Optional axis transpose ----
   # Swap the horizontal and vertical axes.
   if transpose and num_dims == 2:
     values = np.swapaxes(values, 0, 1)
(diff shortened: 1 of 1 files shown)


src/postgkyl/output/plot.py
```diff
@@ -197,6 +197,21 @@ def plot(data: GData | Tuple[list, np.ndarray], args: list = (),
     # end
   # end
 
+    # ---- Optional axis transpose ----
```
manauref: please delete this line. We'd like to keep the style in postgkyl, and not everyone's LLMs add a different style comment. Also, the fact that it's optional is obvious form the signature of the function.

manauref (changes requested):

manauref:
can you add an example?

src/postgkyl/output/plot.py
Antoinehoff: it's not LLM Mana... check other lines, these titles are here.

Antoinehoff:
> can you add an example?

done

src/postgkyl/output/plot.py
Antoinehoff: anyway I removed this one and the others to unify the style.

manauref (approved):

src/postgkyl/output/plot.py
manauref: thanx

PR #227, opened 2026-07-13
add surface plot option and colormap for 1D set of data

Antoinehoff:
We add here the `--surface` option to the `plot` command.

Example:
```
pgkyl tests/test_data/rt_gk_tcv_iwl_1x2v_p1-elc_250.gkyl interp select --z0=2 pl --surface
```
Screenshot 2026-07-13 at 3 51 17 PM

We also make the `--colormap` or `--cmap` option influencing the 1D plots so that it will color the lines according to the passed colormap instead of the usual matplotlib colours. This allows to transform this plot:
```
pgkyl '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-elc_M0_[0-9]*.gkyl' interpolate plot --figure 0 --no-legend
```
image

into
```
pgkyl '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-elc_M0_[0-9]*.gkyl' interpolate plot --figure 0 --no-legend --cmap viridis
```

Screenshot 2026-07-13 at 4 31 24 PM

You can use `cval` to pass an array which will dictates how the different lines maps onto the colorbar, e.g.
```
pgkyl '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-elc_M0_0.gkyl' '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-elc_M0_10.gkyl' interpolate plot --figure 0 --no-legend --cmap viridis --cval 1e-6,1e-5
```
yields

Screenshot 2026-07-13 at 6 45 19 PM

We can then compare the time evolution of the elc and ion M0 moments with
```
pgkyl '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-elc_M0_[0-9]*.gkyl' --tag elc_m0 '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-ion_M0_[0-9]*.gkyl' --tag ion_m0 interpolate collect --use elc_m0 --tag elc_collect collect --use ion_m0 --tag ion_collect activate --tag elc_collect,ion_collect plot --figure 0 --no-legend --surf
```
which yields
image

We can also show it in contour mode (which is the default one)
```
pgkyl '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-elc_M0_[0-9]*.gkyl' --tag elc_m0 '/Users/ahoffman/gkeyll_main/gkeyll/rt_gk_tcv_iwl_1x2v_p1.o/wk//rt_gk_tcv_iwl_csbc_1x2v_p1-ion_M0_[0-9]*.gkyl' --tag ion_m0 interpolate collect --use elc_m0 --tag elc_collect collect --use ion_m0 --tag ion_collect activate --tag elc_collect,ion_collect plot --figure 0 --no-legend
```

Screenshot 2026-07-13 at 5 30 56 PM

### Note
We fix xkcd usage so that it does not contaminate the entire session.


src/postgkyl/output/plot.py
```diff
@@ -1,14 +1,18 @@
 """Module including custom Gkeyll plotting function"""
 from __future__ import annotations
 
+from contextlib import nullcontext
```
manauref, line 4: does this require installing new environment modules?

src/postgkyl/output/plot.py
```diff
@@ -24,6 +28,35 @@ def pgkyl_colorbar(obj, fig : matplotlib.figure.Figure, cax : matplotlib.axes.Ax
   cax2 = divider.append_axes("right", size="3%", pad=0.05)
   return fig.colorbar(obj, cax=cax2, label=label or "", extend=extend)
 
+def get_xkcd_safely(xkcd=True):
```
manauref, line 31: why do we need this? can we get rid of xkcd? it's cute, but it's doesn't really look hand-drawn and it's a pain to keep dragging the code around. If you want a hand-drawn image you can just give chatGPT the non-handrawn version and ask for a hand-drawn-looking one

src/postgkyl/output/plot.py
```diff
@@ -119,411 +165,519 @@ def plot(data: GData | Tuple[list, np.ndarray], args: list = (),
...
+      values = np.swapaxes(values, 0, 1)
+      g0, g1 = grid[1], grid[0]
+      if g0.ndim > 1:
+        g0, g1 = g0.transpose(), g1.transpose()
+      # end
+      grid[0], grid[1] = g0, g1
+      lower[0], lower[1] = lower[1], lower[0]
+      upper[0], upper[1] = upper[1], upper[0]
+      cells[0], cells[1] = cells[1], cells[0]
+      axes_labels[0], axes_labels[1] = axes_labels[1], axes_labels[0]
```
manauref: is this from the transpose PR? please remove

manauref (changes requested):
can you give examples of using
--alpha
--multi2d-mode
--no-multi2d
--cval
too please?

also, i'm surprised it takes +570/-355 line changes to add --surface. It seems to me there are more changes here than necessary (i.e. transpose and xkcd related ones). I think we need to delete xkcd because it's not gonna be used in practice, and it's a drag to maintain that.

Antoinehoff:
> can you give examples of using --alpha --multi2d-mode --no-multi2d --cval too please?
> 
> also, i'm surprised it takes +570/-355 line changes to add --surface. It seems to me there are more changes here than necessary (i.e. transpose and xkcd related ones). I think we need to delete xkcd because it's not gonna be used in practice, and it's a drag to maintain that.

I wanted to push a fix of xkcd also so that it does not contaminate the global variables of matplotlib but is used only locally for the current figure. I don't see a problem with keeping xkcd.

src/postgkyl/output/plot.py
Antoinehoff: I think using liters of water to have a hand drawing of your figure is a very bad usage of ressource. Let's just keep xkcd like that. These changes are fixing some things that were creating warnings (if you don't have the fonts available for example). Now it throw a message on how to get the fonts etc. if you don't have what it needs to use xkcd properly but still runs with what you have.

src/postgkyl/output/plot.py
Antoinehoff: bah

Antoinehoff: just merge transpose, it's ready

src/postgkyl/output/plot.py
Antoinehoff: No this is native python

manauref (approved):
Approving, but I think it'd be good to address the following:
1. I tried this feature but
```
pgkyl <data> <data> interp pl --surfand
pgkyl <data> <data> interp pl --surf --no-multi2d
```
give the exact same plot.

My understanding is that you mean for  `--no-multi2d` to give two separate axes or figures, right? If so, I don't think this flag should exist at all. I think the default is two different figures (as it is with plot already), and if one wants two axes in the same figure one does
`pgkyl <data> <data> interp pl --surf -band`
if you really want to overlay two surfaces you do
`pgkyl <data> <data> interp pl --surf -f0`

2. There are some comments like
```
# ---- Additional Formatting ----------------------------------------
```
and
```
# -----------------------------------------------------------
```
i would remove these. I don't see a precedent for this, but also I see different LLMs adding different things like these, so best to add none (or create a good SKILL file so they all stick to the same style)

PR #228, opened 2026-07-16
Add an interface with the powsqrt operator and new gk quantities.

Antoinehoff:
We add an interface with the powsqrt DG operator (see https://github.com/ammarhakim/gkeyll/pull/1085) and add new quantities:
-sound speed 
-parallel and perpendicular heat fluxs and fluid referential heat fluxe 
-normalized quantities like ephi/T, rho_e/lambda_d, q/nTc_s


src/postgkyl/utils/gk_quantities/registry.py
```diff
@@ -154,6 +154,39 @@
 )
 gk_quant_registry.register(_M2)
 
+# Third parallel velocity moment, int(vpar^3 f) dv.
+_M3par : GkQuantity = GkQuantity(
+  name = "M3par",
+  source = [["M3par"]],
+  fetch_func = [ff.fetch_s0c0],
+  label = r"$M_{3\parallel%s}$ (1/s$^3$)",
```
manauref, line 162: are these units correct? they don't seem like an energy flux to me

src/postgkyl/utils/gk_quantities/registry.py
```diff
@@ -154,6 +154,39 @@
...
+  label = r"$M_{3\perp%s}$ (1/s$^3$)",
+  is_time_dep = True,
+  is_species_dep = True,
+)
+gk_quant_registry.register(_M3perp)
+
+# Third velocity moment, int(vpar*(vpar^2 + vperp^2) f) dv.
+_M3 : GkQuantity = GkQuantity(
+  name = "M3",
+  source = [[_M3par,_M3perp],],
```
manauref: I would add "M3" as the first option, as that file may be available

src/postgkyl/utils/gk_quantities/registry.py
```diff
@@ -246,6 +323,102 @@
 )
 gk_quant_registry.register(_beta)
 
+# Thermal velocity, v_th = sqrt(T/m).
+_vth : GkQuantity = GkQuantity(
+  name = "vth",
+  source = [[_temp],],
+  fetch_func = [ff.fetch_vth],
+  label = r"$v_{th,%s}$ (m/s)",
```
manauref: please remove the h from the label. We have not used it in Gkeyll and we don't wanna start now (partly because when you start working with cross-thermal speeds for collisions then there are too many laters in the subscript).

src/postgkyl/utils/gk_quantities/registry.py
```diff
@@ -154,6 +154,39 @@
 )
 gk_quant_registry.register(_M2)
 
+# Third parallel velocity moment, int(vpar^3 f) dv.
```
manauref: please remove the math from these and other comments describing new quantities in this file. You'll notice I didn't have math in any of the other ones. For the same reason there's no math in metadata: it's hard to have correct math for all quantities, so let's be consistent and have none. We'll add just the necessary math in the fetch_funcs to make the intent clear, and we'll put our energy in placing the math in the gkeyll_gk_quantities overleaf/PDF

src/postgkyl/utils/gk_quantities/registry.py
```diff
@@ -246,6 +323,102 @@
...
+  is_species_dep = False,
+  is_multi_species = True,
+)
+gk_quant_registry.register(_c_s)
+
+# -----------------------------
+# --- Normalized quantities ---
+# -----------------------------
+
+# Square of electron larmor radius over Debye length, gamma parameter in GYRAZE (see eq. 9 of https://arxiv.org/2508.09067).
```
manauref: i'm not convinced we should have normalized quantities. Normalized quantities are simply 2 quantities divided by each other. So I'd think it'd be best to simply call gk-load twice, and divide the loaded quantities. That's more flexible, and doesn't involve creating a growing list of numerator-denominator combinations.

let's discuss offline

what I could do is add weak-operation support to the ev command, so you can do
```
gk-load <quant1_options> gk-load <quant2_options> ev 'f[0] f[1] /'
```
but this would also do any operation between two quantities like add, multiply, sqrt, scale, etc, e.g.
```
gk-load <quant1_options> gk-load <quant2_options> ev 'f[0] f[1] *'
gk-load <quant1_options> gk-load <quant2_options> ev 'f[0] f[1] +'
gk-load <quant1_options> gk-load <quant2_options> ev 'f[0] f[1] pow'
```
and the result would be another DG quantity you can further manipulate/interpolate/plot, etc

If we go this way, you just need to change the entries below so that they instead let us load the numerator an denominator separately, i.e. load
rho_e
lambda_d
n T v_{th} (thermal energy flux?)

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -45,6 +69,28 @@ def _empty_gdata_from_gdata(gdata) -> GData:
...
+def _powsqrt_dg(gdata, exponent: float) -> GData:
+  """
+  pow(sqrt(f), exponent) of a single-component DG field. negative values are set to 1e-40.
+  """
+
+  # We could check negativity before applying the square root.
+  # psi0 = 2.0**(-0.5*gdata.get_num_dims())
+  # if np.any(gdata.get_values()[..., 0:1]*psi0 < 0.0):
+  #   raise ValueError("_powsqrt_dg: the field has negative cell averages, cannot take "
+  #                    "its square root.")
```
manauref: delete
i think the updater sets negative values to 0 already

src/postgkyl/tools/gkeyll_dg_ops.py
```diff
@@ -542,3 +554,57 @@ def invert(self, c_oop: int, oop, c_iop: int, iop) -> None:
...
+    Common exponents: 1 for sqrt(f), -1 for 1/sqrt(f), 3 for f^(3/2).
+
+    Unlike multiply/invert, the gkeyll kernel takes no component index: it
+    reads num_basis coefficients per cell, so both fields must be
+    single-component (scalar). This is checked rather than silently operating
+    on the first component of a multi-component field.
+
+    Note the kernel clamps: at quadrature points where iop is negative it
+    writes 1e-40 instead of failing, so a field that dips negative comes back
+    as ~0 there rather than as an error.
```
manauref: delete. Too verbose.

Comments should say what, not how. The user can refer to gkeyll for additional detail.

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -45,6 +69,28 @@ def _empty_gdata_from_gdata(gdata) -> GData:
...
+  out = _empty_gdata_from_gdata(gdata)
+
+  dgops = GkeyllDGops()
+  dgops.powsqrt(out, gdata, exponent)
+
+  return out
+
+def _sqrt_dg(gdata) -> GData:
+  """Square root of a single-component DG field."""
+  return _powsqrt_dg(gdata, 1.0)
```
manauref: delete, it's an unnecessary indirection. Just call `_powsqrt_dg`

src/postgkyl/utils/gk_quantities/fetch_funcs.py
manauref: delete

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -402,6 +448,264 @@ def fetch_press_p(gdatas, **kwargs):
...
+  one [M0, temp] pair per species, in the order they were requested, e.g.
+    pgkyl gk-load-quantity -q c_s -s elc,ion1,ion2 ...
+  Electrons and ions are told apart by the sign of each species' charge
+  attribute, so the species may be named anything.
+
+  Two definitions are available through '--extra kind=<kind>':
+    ion_acoustic (default): the wave/Bohm-criterion sound speed,
+      c_s = sqrt(T_e*sum_j(n_j*Z_j^2/m_j)/sum_j(n_j*Z_j)).
+    thermo: the bulk-fluid sound speed,
+      c_s = sqrt((gamma_e*n_e*T_e + sum_j(gamma_j*n_j*T_j))/sum_j(n_j*m_j)),
```
manauref, line 703: i think we should make this one the default

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -402,6 +448,264 @@ def fetch_press_p(gdatas, **kwargs):
...
+
+  c_s_sq = _empty_gdata_from_gdata(numer)
+  dgops.multiply(0, c_s_sq, 0, numer, 0, denom_inv)
+
+  return _sqrt_dg(c_s_sq)
+
+_C_S_KINDS = {
+  "ion_acoustic": _c_s_ion_acoustic,
+  "thermo": _c_s_thermo,
+}
```
manauref: place this inside `fetch_c_s`, because I think it's not used elsewhere

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -402,6 +448,264 @@ def fetch_press_p(gdatas, **kwargs):
...
+def _weighted_sum(entries, weights, comp: int):
+  """
+  Sum the comp-th source of each species, each scaled by a scalar weight.
+  """
+  out = _empty_gdata_from_gdata(entries[0]["srcs"][comp])
+  total = sum(w*e["srcs"][comp].get_values() for e, w in zip(entries, weights))
+  out.set_values(total)
+  return out
+
+def _c_s_ion_acoustic(gdatas, **kwargs):
```
manauref: rename `_fetch_c_s_ion_acoustic`, and likewise rename `_c_s_thermo` to `_fetch_c_s_thermo`. This way it's clearer these are sub-types of `fetch_c_s`

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -25,12 +25,36 @@
...
+    if species_idx is None:
+      raise KeyError(f"fetch function: '--extra {key}=' was given {len(val)} values but this "
+                     f"quantity is not computed per species, so there is no way to tell which "
+                     f"one to use. Pass a single value instead.")
+
+    if species_idx >= len(val):
+      species = kwargs.get("species", None)
+      raise ValueError(f"fetch function: '--extra {key}=' was given only {len(val)} values but "
+                       f"species #{species_idx}{f' ({species})' if species else ''} was requested. "
+                       f"Give one value per species, in the order of '--species'.")
```
manauref, line 48: this stuff is more complicated just because of working with old data files that don't have metadata. We should not add more code to support old data files. Much of what we've done in postgkyl recently hinges on having the metadata. So please undo this. The mass/charge should just come from metadata.

If you want you can push this bit to a branch so that someone working with old data who wants to use new features can use that, but I suspect other stuff (like gk-nodes) will also not work with old data. We can't make all the new stuff backward compatible.

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -73,15 +119,15 @@ def fetch(gdatas, **kwargs):
...
       raise ValueError(f"Datasets have different basis")
 
-    vals_l = gd_l.get_values()
-    vals_r = gd_r.get_values()
-  
-    out = GData(ctx=gdl.ctx)
+    vals_l = gd_l.get_values()[..., ck*nb_l:(ck+1)*nb_l]
+    vals_r = gd_r.get_values()[..., cl*nb_r:(cl+1)*nb_r]
+
+    out = GData(ctx=gd_l.ctx)
```
manauref, line 115: uh, these seem like big bugs. I'm surprised we didn't see this affect results before. Glad you caught it

src/postgkyl/utils/gk_quantities/fetch_funcs.py
```diff
@@ -402,6 +448,264 @@ def fetch_press_p(gdatas, **kwargs):
...
+  Split the per-species sources of a multi-species quantity into the electron
+  entry and the ion entries, by the sign of each species' charge..
+  """
+  species_names = kwargs.get("species", [])
+  if len(species_names) != len(gdatas):
+    species_names = [f"#{i}" for i in range(len(gdatas))]
+
+  elcs, ions = [], []
+  for species_idx, (name, srcs) in enumerate(zip(species_names, gdatas)):
+    # Resolve each species' attributes against its own slot in a '--extra' array.
```
manauref, line 599: see my comment about not using --extra for data in metadata. I think it'll simplify this bit.

src/postgkyl/commands/gk_load_quantity.py
```diff
@@ -22,7 +77,9 @@
 @click.option("--label", "-l", default=None, type=click.STRING,
   help="Label override for the output dataset.")
 @click.option("--extra", "-e", default=None, type=click.STRING,
-  help="Extra comma-separated key=value pairs of extra commands, e.g. dir=1,mass=0.1. Purpose depends on -q.")
+  help="Extra comma-separated key=value pairs of extra commands, e.g. dir=1,mass=0.1. "
+       "A key may be given one value per species as a comma-separated array, e.g. "
+       "mass=me,mi1,mi2 alongside --species elc,ion1,ion2. Purpose depends on -q.")
```
manauref, line 31: delete. Should just use values from metadata

manauref (changes requested):
See comments below. I think the main big picture comments are:
1. I don't think we should add new code to support old files without metadata. We are at a point when anyone adopting these new tools can run sims with the latest gkeyll that has the meta.
2. I don't think we should add normalized quantities to the registry. It is better to: A) add the numerator and denominator as separate quantities, B) empower `ev` to do weak operations. This lets people normalize any quantity as they wish (e.g. what if I want to normalize qpar to (n T c_s) instead of (n T v_t)?). I can help with changes to `ev`

src/postgkyl/utils/gk_quantities/registry.py
Antoinehoff: Since M2 is in m^-1 s^-2, I assumed that [M3] =[v] * [M2] = s^-3. I think it's correct because w/m^2 = kg/s^3 and we have to multiply M3 by the mass to get the heat flux.

Antoinehoff:
@manauref I think this `--extra` should stay. We can remove it in another PR. For example, the `--extra dir=1` is bad. When we load a vector, we should just load each component and then use select -c 1 to have only one component.

src/postgkyl/utils/gk_quantities/registry.py
Antoinehoff: hm ok I will keep these in my own branch because I may use them a lot and I want to avoid command and code bloating.

Antoinehoff:
another point @manauref, I think in general we have to different approaches: you don't want to bloat the code inside the code base while I don't want to bloat the interface... I wish there would be a perfect world but it feels like these two are a bit in contradiction with how postgkyl is written.

manauref (approved):
Approving but please make these changes:
1. Delete the `personal/` directory. This will lead to people just developing their own stuff instead of the shared tool, just because they want to avoid conflict.
2. I changed my mind about normalized quantities. Put them back in the registry BUT make sure the numerator and denominator are themselves independent, loadable quantities.

In the future we may find an equally easy place to do normalizations, for which we'll need the numerator and denominator separately.

PR #229, opened 2026-07-17
Refactor of postgkyl

### LINKED ISSUE #6: Refactor the outdated DG part of the code

### LINKED ISSUE #7: Alternative interpolation points
Would like ability to interpolate onto non-equidistant nodes. For example, would like to interpolate onto x = +/- 1/3 in each [-1,1] cell, or onto quadrature nodes x = +/- 1/sqrt(3).

### LINKED ISSUE #13: Weak operations in postgkyl
Weak division, multiplication, and other vector operations (dot product and cross product) should be possible on the DG fields postgkyl reads in. 

This feature will allow us to compute weak moments (such as u, T, P_{ij}, q_i, the flow, temperature, pressure tensor, and heat-flux vector) even when we do not write out weak moments (and in general this may be preferable to writing out large numbers of files). In addition, there are a number of operations we would like to do with Vlasov (and eventually DG fluid) simulations that can also be done weakly, thus reducing error propagation. For example, in drift analysis and energization analysis, we are often interested in the parallel/perpendicular components of a field. The vector form of the parallel flow is simply:

u_{parallel} = (u dot b_hat) b_hat

which can be expressed as:

u_{parallel} = [(u dot B) * B]/|B|^2

the entire operation of which can be done using weak multiplies, dot products, and divides. We can then find the total parallel energy gain by doing a similar set of weak multiplies, dot products, and divides on the electric field, and computing: u_{parallel} dot E_{parallel}.

Other manipulations of the data, such as the parallel pressure, or E x B velocity, can also leverage weak operations, allowing us to delay interpolating of the data until the moment before visualization in many cases.

To add this feature, I believe the quickest thing to do would be to copy BinOp kernels into the postgkyl repo with a similar organization to gkyl, and then use similar infrastructure to the interpolate method to manipulate the raw DG field. Alternatively we could work towards making gkyl directly callable by postgkyl, which definitely speed up the operation (though at least for the application I want, this feature is only needed on 1D, 2D, and 3D data, and I expect that looping over a configuration space grid and doing weak operations is not going to be that taxing).

### LINKED ISSUE #52: Add support for hybrid basis used in Vlasov simulations
In [gkylzero](https://github.com/ammarhakim/gkylzero) we've defaulted polyOrder=1 in Vlasov simulations to mean hybrid basis, meaning p=1 in position and p=2 in velocity space. Currently pgkyl doesn't know how to handle ```hybrid``` basis. The problem is that in various places we make decision based on polyOrder or dimensionality, but this has a mixed order, and dimensionality alone doesn't determine the number of basis functions; for example

1x3v
2x2v

both have ndim=4, but they have different numbers of basis monomials. It seems like one would have to make several places aware of cdim and vdim to make this work (or there may be other ways). This is not a tremendous priority because most Vlasov users don't run with p=1/hybrid (although there may be some benefits to doing so in some cases?).

On the other hand there is already support for ```gkhybrid``` basis because in that case ndim uniquely specifies the basis (i.e. 1x1v,1x2v,2x2v,3x2v all have a distinct ndim).

### LINKED ISSUE #108: Allow specifing interpolation points in interp command
At present pgkyl chooses the interpolation points in a cell based on polyOrder. However, it would be useful to allow the user to specific the interpolation points in the reference hypercube [-1,1]^d. This would allow plotting points at user controlled locations, making some sorts of tests and plots much easier to do.

### LINKED ISSUE #141: Hybrid basis specification in interp
I thought we replaced the pkpmhyb and gkhyb with just hybrid? However, I get the following:

```
gandalf@valinor ~/r/g/g/m/gkylzero> pgkyl interp -h                                                 (pgkyl) 
Usage: pgkyl interp [OPTIONS]

  Interpolate DG data onto a uniform mesh.

Options:
  -b, --basis_type [ms|ns|mo|mt|gkhyb|pkpmhyb]
                                  Specify DG basis.
  -p, --poly_order INTEGER        Specify polynomial order.
  -i, --interp INTEGER            Interpolation onto a general mesh of
                                  specified amount.
  -u, --use TEXT                  Specify a 'tag' to apply to (default all
                                  tags).
  -t, --tag TEXT                  Optional tag for the resulting array
  -l, --label TEXT                Custom label for the result
  -r, --read BOOLEAN              Read from general interpolation file.
  -h, --help                      Show this message and exit.

```

### LINKED ISSUE #159: Plotting grids in 2D and 3D
We need a way to visualize 2D and 3D grids. The grids can be either specified as a NDIM component field with coordinates for each node specified, or a mapc2p representation as a DG field (which may be higher than just piecewise linear).

This is important to be able to viz the grids we make for moments as well and for the DG code, including GK. Eventually we want to support 1D, 2D and 3D grids. 

To get the process moving I have created a 2D grid viz command called `showgrid` in the `plot-grid` branch. For now this only works in 2D and does no error checking at all. However, it illustrates the method and can be used a starting point for making more production quality plots with titles, labels, etc. For now it allows to make plots of 2D grids so I can viz them. But we need to improve with more features.

I have added the test data file `tests/test_data/nodal_coords-sqrt-psi.gkyl`. The command:

`pgkyl nodal_coords-sqrt-psi.gkyl showgrid --aspect 0.05 
`

will make the following plot:

Image

### LINKED ISSUE #188: Plot saving also saves an unnamed .png file
I've noticed that there are these stray `.png` files without any surname. This happens any time I save data from command line.

### LINKED ISSUE #198: [DR] Change command "pr" to "print"
We should allow users to type pgkyl <data> pr, but the command should be print so that the name is clear regarding what its doing.

### LINKED ISSUE #210: [DR] API for scripting interface
I'm reopening this DR after discussing with Ammar that the key issue with the one before was the direct inclusion of Claude's output in the text. This text is entirely human-written, with Claude's opinion in a file, if you're interested.

I was inspired by a YouTube video about readable code.
https://m.youtube.com/watch?v=SJocPm2E8eQ
Part of my most recent frustration comes from the disconnect between the script pgkyl and the command-line pgkyl, which have very different interfaces. This leads people to align with either one or the other. It would be great to have a more command-line-like interface we could use in pgkyl, similar to this. In contrast to my earlier DRs, I'm suggesting thinking about the script mode pgkyl from a top-down approach. It will benefit us to think about the final scripts we want to write, rather than an API-first approach. I know this is a standard goal, ideal, and objective for our projects, but I don't think it's been applied well to script pgkyl. 

In a Python script, we could plot some data like

```python
import postgkeyll as pg

pg.load('filename').select(z0=0.0).plot()
```

This could also write the common data object to avoid using activate and tagging.

```python
import postgkeyll as pg

data1 = pg.load('filename').select(z0=0.0)
data2 = pg.load('file2').evaluate('f 5 +').select(z0=0.0)
pg.plot(data1.and(data2))
pg.print(data1)
```

Here, the `and()` method simply adds data1 and data2 to the same stack for pg plotting.

The goal is to construct a "language" for how to write pgkyl scripts, similar to the YouTube video describes. In practice, this could be an object-oriented wrapper for our underlying functionality.

We can define a dunder method for these data classes to perform regular Python operations on these objects (print, add, multiply, divide). This would make manipulation more intuitive.

Additionally, we can use this interface to interact with existing structures like numpy. We can (and should) put guardrails on these methods so that we can't perform NumPy operations on DG non-interpolated data.

```python
import postgkeyll as pg
import numpy as np
import matplotlib as plt

a = pg.load('file_a').interp()
b = pg.load('file_b').interp()
c = np.sqrt(a**2 + b**2)
plt.plot(c.grid, c)
```

Here is what Claude came up with for a plan of structuring the API. It seems interesting.

[READABLE_API.md.rtf](https://github.com/user-attachments/files/29266719/READABLE_API.md.rtf)

These are just ideas for a scripting interface, and I'm very open to ideas contributing to this script-first approach.

The interface with the CLI is very, very simple. The driver of Postgkeyll, in this case, is a master class that has methods for the commands. The commands can be called from these objects. The CLI wraps the master object using click (I learned recently that `Typer` is the modernized version of click). So the layer here sits between the commands/ and click, while the current structure has the commands orchestrated with click.

Additionally, this master class should be chock-full of examples that are verified through CI. This way, it's very obvious how to use the package.

### LINKED ISSUE #218: [DR] Separating loading data from non-uniform mappings
The procedure for loading data into memory and mapping from uniform to non-uniform coordinates is presented as two distinct operations on the data; however, the current codebase structure groups them together. While I see that performing a map on the data loads a file, it is fundamentally an operation on the grid; therefore, this is an operation, not a loading procedure. These two operations can be separated.

After chatting with Claude, I realize that the current implementation is actually misleading. The grid mapping actually doesn't happen until during interpolation for DG data. It's misleading to say the mapping operation happens during loading, because it actually occurs during interpolation. This mapping should ideally be separated out as its own distinct method and operation, rather than grouped in with these other operations.

### LINKED ISSUE #221: Rename write to save
It is most cohesive and understandable to mirror the numpy API with gkeyll functions from a users perspective. In this light, numpy uses variables load and save. https://numpy.org/doc/2.3/reference/generated/numpy.load.html. We use write, not save. It would make more sense in the python ecosystem to use a name "save" instead of "write" for saving a .gkyl file.

Maxwell-Rosen:
# Refactor of postgkyl

This pull request addresses pain points in postgkyl development. First, postgkyl currently has its own sympy-generated copy of DG kernels that Gkeyll already implements in C. Second, it is a CLI-first library and an awkward scripting interface. Third, equation-informed commands are mixed with basic operations, making structure unclear. This PR fixes all three at once: all DG algebra now runs through Gkeyll’s own compiled kernels via a shim that lives in the gkeyll tree (so drift is a compile error). A fluent Python API becomes the primary surface through GData, with the CLI as a thin shell over it. The new hierarchy separates core operations from equation-informed diagnostics, with the separation enforced by a unit test. Unit testing coverage rises from 33% to 99%. These improvements to the codebase will make postgkyl easier to use, simpler to maintain, and straightforward to extend.

## Related GitHub issues

Closes #221. Rename write to save. Performed in this refactor.
Closes #218. [DR] Separating loading data from non-uniform mappings. A verb `map` is introduced so that loading simply loads data.
Closes #210. [DR] API for scripting interface. This PR implements the API, therefore closing the design review.
closes #198. [DR] Change command “pr” to “print”. All commands are given their full name. The Python side does not use abbreviations or shorthand (interp is now interpolate). The CLI always performs minimal string parsing so that pr resolves to print; unless prandtl is also a command, you have to specify pra/pri.
Closes #188. Plot saving also saves an unnamed .png file. This PR fixes this issue. I’ve fixed this many times in branches.
Closes #159. Plotting grids in 2D and 3D. This PR includes plotly and pyvista inside. This PR was approved by Ammar, but is still blocked by Mana.
Closes #141—hybrid basis specification in interp. We add support for the hybrid basis via the Gkeyll C library in this PR, closing this issue.
Closes #108. Allow specifying interpolation points in interp command. This is addressed using Gkeyll’s quadrature points and nodal evaluation of the interp command. This is a different statement that allows users to specify exactly the logical coordinates at which they want points evaluated.
Closes #52. Add support for hybrid basis used in Vlasov simulations. The Gkeyll C library supports this; therefore, closing this issue.
Closes #13. Weak operations in postgkeyll. This PR supports weak division, multiplication, and integration, with many more features likely to be developed.
Closes #7. Alternative interpolation points. This PR supports quadrature node interpolation and cell edges, resolving this issue.
Closes #6. Refactor the outdated DG part of the code. Oldest issue in the codebase. This issue is resolved by using the C layer of Gkeyll to perform DG operations, utilizing the maxima auto-generated code, and unifying the codebases.

## Gkeyll backend support

A linker `gpython.c` is inside `gkeyll/core`, which exposes operations such as array_ops and integration to Python. Postgkeyll has a sparse git submodule for `gkeyll/core`, which builds a dependency-free version of postgkyl at pip install time. Gkeyll’s functionality is linked through an internal compiled file. This design circumvents the need for `ctypes` or `cffi`. This structure is advantageous because if the wrapper becomes out of sync, BOTH postgkyl and gkeyll will error, so gkeyll will fail to compile. These errors are loud and diagnosable, while `cffi` and `ctypes` can silently fail. This installation works out of the box with pip, without requiring users to install Gkeyll themselves first. For systems where Gkeyll cannot compile, a pure-Python reader is available. Similar to the Lua-C API rewrite, this hardens the compatibility layer and helps with maintainability.

### Modal versus nodal operations

Gkeyll owns everything modal, as this functionality is precisely what the Gkeyll repository is built for. Gkeyll performs weak arithmetic, integration, interpolation, and DG evaluation. On the other hand, pointwise operations are exact, so NumPy is responsible for these operations. NumPy affords a rich ecosystem of commands, affording great flexibility. Only interpolation crosses the line between Gkeyll arrays and NumPy arrays. `.gkyl` files default to loading as modal, but sometimes we store nodal data. As a result, we must store whether an array is modal or nodal in the metadata of the files we save.

## Unified script / CLI interface

The script mode and CLI interface now mirror each other. In script mode, one can type

```python
import postgkyl as pg

pg.load('file').interpolate().select(z0=0.0).plot()
```

While in the command line, one types

```bash
pgkyl file interp sel --z0 0.0 pl
```

Having a unified syntax for both the script and the CLI simplifies working with this package and makes post-processing review easier.

Scripts can go even further, and here is the syntax that this PR aims to support and extend script mode.

```python
import numpy as np
import postgkyl as pg

m0 = pg.load("sim_elc_M0_5.gkyl")
m1 = pg.load("sim_elc_M1_5.gkyl")
upar = m1 / m0              # weak division, inside Gkeyll's kernels
quad = upar.to_quad()       # values at quadrature points
root = np.sqrt(quad)        # NumPy ufuncs work on point-value data
modal = root.to_modal()     # exact projection back to DG coefficients
print(modal)                # prints like a NumPy array
```

To keep script and CLI in sync, both call upon the same API. The click interface is a thin layer on top of the API, making the repository more modular, while keeping all aspects in sync.

### A necessary split

There is a fundamental issue with using this command structure. `GData` presently contains the grid and values, but introducing methods creates a paradox. The operations need to import `GData`, but `GData` has methods of operations, forming a cycle. To break this cycle, `GData` is separated into 2 classes, `GData` and `GDataState`. `GData` contains the methods, while `GDataState` contains the data itself. These objects live in different layers of the hierarchy.

### Unit testing

postgkyl on main only has 33% coverage. This refactor was meticulous in its use of test-driven development. Now, the repo has 99% code coverage, including our most fundamental functionality. The layer hierarchy is enforced by the `test_postgkyl.py` test, which walks through every real import and checks it against an allowlist. The architecture diagram cannot silently rot because a test fails when a test fails.

## The postgkyl hierarchy

Similar to Gkeyll, this PR implements this proposed hierarchy. Having layers in postgkyl keeps our code organized, modular, and readable. This layering mirrors the Gkeyll repository.

### What are diagnostics

This PR separates commands into a set of `operations/` and `diagnostics/`. Operations are fundamental low-layer manipulations of the data (e.g. `interpolate`, `plot`, `select`), without any knowledge about the data they are operating on. Diagnostics are equation-informed tools which help users understand the Gkeyll simulations (e.g. `10m`, `gyrokinetics`, and `vlasov`). Sitting on top of the API, they can call the fundamental operations, just as a Python script would. This layer is highly flexible, enabling future users to implement and share their own post-processing tools without worrying about the lower levels of the codebase. While the CLI must group commands to preserve chaining, refactoring it into a thin wrapper affords greater modularity for organizing the underlying scripts.

### File tree

Below is the file tree, in hierarchical order. This hierarchy is enforced by a unit test `test_postgkyl.py`.

```
src/postgkyl/
│
├─ __init__.py          facade · `import postgkyl as pg`              [SURFACE]
│
├─ cli/                 Thin Click shells: argv → gdata / diagnostics   [SURFACE]
│   ├─ app.py
│   └─ commands/        All the callable commands from the command line
│
├─ diagnostics/        equation-specific physics · one module        [COMPOSITION]
│                      per equation model
│
├─ gdata/                ★ THE FLUENT SURFACE  (sits ABOVE operations)  [FLUENT API]
│   ├─ gdata.py          class GData(GDataState) + .interpolate()/.plot()
│   ├─ gdatagroup.py     fluent GDataGroup: broadcasts verbs over its members
│   ├─ verbs.py          module-level fluent verbs with no single `self`
│   │                    (collect/evaluate/relchange/animate) — one-line
│   │                    delegations to `operations`, shared by GData and GDataGroup
│   └─ load.py           pg.load(...) → returns a GData
│
├─ operations/         one function per verb · the single seam        [VERBS]
│   ├─ interpolate.py    interpolate(d: GDataState) -> GDataState      (core verbs
│   ├─ select.py                                                       only —
│   ├─ plot.py                                                         equation-blind)
│   ├─ animate.py        terminal: sequence of datasets → render's animation engine
│   ├─ average.py        terminal-adjacent: weighted average over a dim subset,
│   │                    stays modal/gkyl-native (composes with further verbs)
│   ├─ eval_at_coord_proj.py  terminal-adjacent: eval at coords, project to the
│   │                    lower-dim basis for survivors, stays modal/gkyl-native
│   ├─ local_poly.py     modal coefficients → discontinuity-preserving plot mesh
│   └─ _materialize.py   shared modal → NumPy-shadow bridge used by plot/animate
│
├─ render/             matplotlib · plotly · pyvista → gdatastate/numerics [BACKEND]
│                      (below operations, which delegates plot() to it)
│
├─ gdatastate/         ★ THE CONTAINER  (state only, NO verbs)        [CONTAINER]
│   ├─ gdatastate.py          class GDataState: grid·values·ctx·_result·dunders
│   ├─ gdatastategroup.py     GDataStateGroup
│   └─ guards.py         shared field-domain guard (backend=="gkyl" -> raise);
│                        one home for the ".interpolate() first" check reused
│                        across operations/diagnostics instead of retyped per verb
│
├─ numerics/           pure NumPy math · 0 internal imports           [LEAF]
├─ dg/                 interpolation bridge + modal ops → gpython     [ENGINE]
├─ io/                 readers (C-native first) + writer → gpython    [ENGINE]
└─ gpython/            ★ THE FOREIGN FLOOR · compiled shim            [FLOOR]
    ├─ csrc/             _gpythonmodule.c — CPython extension over gkyl_gpython.h
    │                    (the shim itself lives in gkeyll/core/zero/)
    ├─ _gpython.so       built extension (scripts/build_gpython.sh)
    ├─ _lib.py           loads _gpython · GPYTHON_API_VERSION handshake
    ├─ array.py          GkylArray — capsule owner of a gkyl_array
    ├─ basis.py          basis cache + interpolation matrices via the shim
    ├─ rio.py            file loading via gkyl_array_rio
    └─ kernels.py        weak mul/div/inv · lincomb · reduce · integrate
```

A graphical user interface (GUI) can operate at the surface level, near the CLI, earning the abstraction of another interface.

### Version-breaking changes

Since this PR is an entire rewrite of the library, I anticipate it will have version-breaking changes. The old GData class has a fundamentally different structure. Please reach out if your old scripts no longer work and we can fix them. I am very open to feedback that improves the design of our codebase.

I paid special attention to porting the latest developments for gyrokinetics postprocessing, but this took several weeks, and the codebase is very active, so there is a non-negligible chance that I missed something. The best way to know is to get others' eyes on the code and use it.

Nodal arrays need a property `value_forrm = "nodal"` added to their metadata. Here is an example.

```c
  struct gkyl_msgpack_map_elem meta[] = {
    { .key = "poly_order", .elem_type = GKYL_MP_UNSIGNED_INT, .uval = 0 },
    { .key = "basis_type", .elem_type = GKYL_MP_STRING, .cval = "serendipity" },
    { .key = "value_form", .elem_type = GKYL_MP_STRING, .cval = "nodal" },
  };
```

### Agentic development

This refactor was carried out by Claude Code. I constructed a 15-staged refactoring plan with associated agents to orchestrate. I have attached the files for consideration. The agent was created in Fable 5, reviewed and then implemented with Sonnet 5. Correctness is enforced through a comprehensive layer of unit testing and human review.

### Bug fixes

Along the way, the agents found several (5-10) genuine bugs in the codebase. They can be found in the agent's reports and the commit history. These bugs are fixed by the agents, and unit tests are added.

### Known limitations

This refactor brings out some limitations of Gkeyll itself. Gkeyll does not have div/mul/integrate kernels for gkhybrid and hybrid basis functions. Integration for the GK distribution function is in gyrokinetic/ rather than core/. This is beyond the scope of this PR.

Modal integration is only supported in total. Per-axis integration was only supported in a nodal since in postgkyl, which is carried over. Gkeyll does not support per-axis integration in the kernels, so Gkeyll must be extended for this application.

### Out of scope

In a perfect codebase, I'd like to have the website compiled directly from postgkyl, linked to tested examples and checked through CI. The codebase should also include linting and formatting, but I didn't implement them here. That's a separate discussion. I've included a few walkthrough scripts to guide new users through the codebase. We should prevent people from putting code in main that lacks unit tests. CI should check code coverage to keep our coverage high.


manauref (changes requested):
Good work Max. This is a compelling refactor.

I haven't gone through nearly as much of the code as I'd like before we merge (though I probably won't go through all). But I'd like to make 2 initial requests before this PR matures further:

1. We should take advantage of this refactor to improve handling of multiblock data. Presently, main does not handle multiblock data well. Some features are not supported (e.g. I doubt`gk-rz` works with multiblock), some commands support it one way while others support it another way (e.g. gk-nodes, plot, animate). This is because multiblock support in postgkyl has been done in a piecemeal way, rather than designing its support taking into account the whole workflow (loading, processing, visualizing). Now that the workflow is being further modularized and improved, we should address this flaw.

Here's an idea:

Just as gkeyll's multiblock solver is a group of single-block solvers, perhaps there should be a multib layer in postgkyl which simply loops over blocks and does all the (existent)  (loading, processing) operations to each block, and then visualiztion (plot, animate) just groups them onto the same figure. For example, the cli could be something like
```
pgkyl multib <sim_name>-b[0-9]*_elc_M0_1.gkyl interp gk-rz pl
pgkyl -m <sim_name>-b[0-9]*_elc_M0_1.gkyl interp gk-rz pl
```
or simply
```
pgkyl <sim_name>-b[0-9]*_elc_M0_1.gkyl interp gk-rz pl
```
(note that I moved the _b# in the name to the other side of the `-`, which will make the parsing easier)
then internally, it would loop through the blocks and do loading+interp+gk-rz to all the data, and pl would simply push the to the same figure/axis. This way none of the operations need to be multiblock aware and need hardly any (or no) modifications. Only the visualization commands (e.g. pl, animate) need to be multiblock aware.

Some (perhaps future) commands would need to combine data from blocks, e.g. to perform reductions across blocks. For example, the `dg_integrate` or `dg_average` commands may eventually need to have additional cross-block functionality incase one wants to, for example, average across all blocks or across x of x-connected blocks. This would also mirror gkeyll's multiblock solver, which mostly loops over blocks and performs single-block operations, but there are some *select* inter-block operations like reductions and Poisson solvers. We can worry about this later.

We can/should add the block ID to the metadata, and not just rely on the file name.

2. Please make sure that recent features that @Antoinehoff and I have added continue to work. Specifically i'm thinking of:
- gk-rz (interpolates DG data onto the R-Z plane).
- gk-load
- dg-local-poly
- evaluation of a DG dataset at a specific coordinate and projection onto a lower dimensional basis
- dg-avg
- dg sqrt 
- `--surface` option in plot

we can provide sample cli/script examples of each.

Maxwell-Rosen:
I think these comments about the multiblock handling should be a separate issue. I would say this exceeds the scope of this PR, which is intended to rewrite existing functionality, not add new functionality. We can address this as well, but we should first review the library's fundamentals. As I don't deal with multiblock data, I don't have a clear understanding of how users interact with it. It's not difficult to add with this refactored version, but we shouldn't block this major refactor behind a specific feature related to a subset of use cases. This refactor handles multi-block simulations the same way as postgkyl 1.0, so it should not block the merge.

All merged PRs, as of today, are present in this branch, including the ones you've listed.

README.md
```diff
@@ -13,25 +13,14 @@ Full documentation of the Gkeyll project is available at
 
 ## Dependencies and Installation
 
-Postgkyl requires the following packages:
+Postgkyl requires the packages listed in pyproject.toml
```
manauref: The installation instructions are, overall, confusing. It needs to be written so a high-schooler can just follow a set of steps and do it, without needing to know what a number of things you mention are.

I suggest you rewrite it with this structure:
```
## Installation

1. Create environment
(at least mention conda, even if you are not going to provide detail steps)
### pyenv
### mamba
2. Install dependencies
distinguish between dependencies in environment.yml and those in the .toml file (not sure why they can't just stay in a single file, but ok)
3. Install postgkyl

## Documentation


### Additional installation notes
#### Gkeyll bridge

## Developing for postgkyl

### Running tests
pytest dependecy and how to run tests
### formatting
```

This way the user is first faced with just what it needs to know (installation+documentation). The rest is to be left for later sections.

Maxwell-Rosen: addressed

PR #230, opened 2026-08-06
remove the Monty Python ref from the print verbosity

Antoinehoff:
When you find yourself using verbose mode, you are not in the mood to get a joke at each postgkyl call :(


manauref (approved):
gotta make sure new PRs get merged into @Maxwell-Rosen 's refactor

